PR #12, opened 2019-10-30
add overides to platform

maxandersen:
platform MUST have categories so adding the override file here.

should we delete the one in core ?


gastaldi:
Shall we merge this and release 0.0.3?

PR #22, opened 2020-01-09
Decrease importance of Core category

emmanuelbernard:
Also force YAML and JSON as pinned.

PR #24, opened 2020-01-29
remove camel extensions override as recent camel has this in place now

maxandersen:
as mentioned in #22


aloubyansky:
Looks like bad formatting. Extra `]` at the end?

maxandersen:
I could have sworn I built it without error, I guess not :) fixed.

maxandersen:
fixed and this is now a blocker for 1.3 as the version numbers in here messes up the output

PR #28, opened 2020-02-26
add Qpid JMS extension

gemmellr:
Add a Qpid JMS extension, from https://github.com/amqphub/quarkus-qpid-jms, for use in listing on the code.quarkus.io website and also in an updated JMS guide (changes still to be done).

The PR is based on the current 0.11.2 extension release which uses Quarkus 1.2.1.Final. I have the extension master updated for 1.3.0.Alpha2, and a nightly build running against Quarkus master also ahead of the upcoming CR.

It was suggested I actually raise a PR for adding the extension based on using 1.3.0.Alpha2, however I found that due to the changes between 1.2.1 and 1.3.0 either the build/tests for this extension or those for some other existing camel extension modules would fail if doing that (depending on whether or not I updated the version the platform pom was using). I decided to raise the PR based on the existing 1.2.1 as a result in order to get feedback on the functional build. I can then make any changes needed based on this, and create an updated version based on either 1.3.0 Alpha2 or CR as appropriate.


gemmellr:
I notice other PRs have reviewers set but I dont seem to be able to add any so I am just going with a basic mention in case you haven't seen it yet: @aloubyansky @gsmet  could you give this a look over please?

Thanks,
Robbie.

bom/deployment/pom.xml
```diff
@@ -43,6 +43,14 @@
...
             </dependency>
 
+            <!-- Qpid JMS -->
+            <dependency>
+                <groupId>org.amqphub.quarkus</groupId>
+                <artifactId>quarkus-qpid-jms-parent</artifactId>
+                <version>${quarkus-qpid-jms.version}</version>
+                <type>pom</type>
+                <scope>import</scope>
+            </dependency>
```
aloubyansky: 1) Given that this BOM imports the runtime one which in its turn imports `quarkus-qpid-jms-parent`, this import is not effective.
2) In Quarkus we differentiate between "runtime" and "deployment" dependencies. What does `quarkus-qpid-jms-parent` contain?

gemmellr: The quarkus-qpid-jms-parent currently has the qpid-jms-client, the quarkus-qpid-jms and quarkus-qpid-jms-deployment modules for the extension, and quarkus-build-parent: https://github.com/amqphub/quarkus-qpid-jms/blob/0.11.2/pom.xml#L91-L120

I mainly copied that from the camel example above, but I admit I didn't pay close attention to what that contains.

aloubyansky: If you do have a runtime and deployment BOM then please import your runtime BOM into the runtime platform BOM and the deployment one into the deployment platform BOM.

gemmellr: The extension doesn't have boms, the referenced https://github.com/amqphub/quarkus-qpid-jms/blob/0.11.2/pom.xml#L91-L120 is where the versions are managed currently.

The dependencies except the client jar and extension jars are already (or will be in 1.3.0) managed by the core Quarkus bits, if that matters.

aloubyansky: Ok, `quarkus-build-parent` should not be imported in neither of the platform BOM.
Just import
```
    <dependency>
        <groupId>org.amqphub.quarkus</groupId>
        <artifactId>quarkus-qpid-jms</artifactId>
        <version>YOUR_VERSION</version>
      </dependency>
```
in the runtime platform BOM. And
```
      <dependency>
        <groupId>org.amqphub.quarkus</groupId>
        <artifactId>quarkus-qpid-jms-deployment</artifactId>
        <version>YOUR_VERSION</version>
      </dependency>
```
in the deployment platform BOM.
And I guess you'll want the client in the runtime, right?

gemmellr: Ok I can change that. To add the client to the platform runtime bom explicitly I would need a further version property, with both needing changed for updates, is that ok?

aloubyansky: Either that or create your runtime BOM and import it into the runtime platform BOM (perhaps that would look simpler and may turn out to be a good practice for every contributor). The maintenance of that will be your (team's) responsibility.

gemmellr: I've pushed an update reflecting the original suggestions since that works with whats available now. I'll be the one raising updates anyway, and I'd still need to figure out what would go into a bom and create+release it, which can still come later.

bom/runtime/pom.xml
```diff
@@ -34,6 +34,20 @@
...
+                <groupId>org.amqphub.quarkus</groupId>
+                <artifactId>quarkus-qpid-jms</artifactId>
+                <version>${quarkus-qpid-jms.version}</version>
+                <type>pom</type>
+                <scope>import</scope>
+            </dependency>
+            <dependency>
+                <groupId>org.apache.qpid</groupId>
+                <artifactId>qpid-jms-client</artifactId>
+                <version>${qpid-jms.version}</version>
```
gsmet: I don't understand why this isn't in your BOM?

gemmellr: As it isnt a bom, and I had only partially made the feedback change I was meant to, forgetting to remove the import. I have pushed it again.

From earlier comments when I was importing quarkus-qpid-jms-parent:
> The extension doesn't have boms, the referenced https://github.com/amqphub/quarkus-qpid-jms/blob/0.11.2/pom.xml#L91-L120 is where the versions are managed currently.
> 

For extra context, I based what quarkus-qpid-jms-parent does on what the extensions in Quarkus were doing when it was written and appear to still do, where they inherit quarkus-build-parent at their own parent pom and dont have any dep management of their own in runtime or deployment poms, so I guess they are getting the same dep management in both cases.

aloubyansky: User applications will import only the runtime part though.

gemmellr:
I have created an initial change for the extension to add and use a bom:
https://github.com/gemmellr/quarkus-qpid-jms/compare/0.11.x...gemmellr:bom

Plus an updated version of the platform changes using the bom:
https://github.com/gemmellr/quarkus-platform/compare/master...gemmellr:qpid-jms-extension-bom

Can you give those a look over? If they are suitable I can go ahead and make the extension change, release it, and finally update this PR.

gemmellr:
The main thing I wondered on was whether the quarkus-qpid-jms-bom bom should also import quarkus-bom?

aloubyansky:
> The main thing I wondered on was whether the quarkus-qpid-jms-bom bom should also import quarkus-bom?

I'd say so. The way you did it makes your BOM dominate over the `quarkus-bom`, which could be ok until you try and integrate it into the platform where we make sure `quarkus-bom` dominates over the integrated extensions. So I think if you were following the same model for your extension, the integration into the platform would hopefully have less surprises.

gemmellr:
Ok. I've imported `quarkus-bom` first, the same way the platform does. If that looks good I'll go ahead and release then update the PR.

aloubyansky:
Looks good @gemmellr

gemmellr:
Great, thanks. I'll push it up properly and release the extension shortly.

gemmellr:
Actually, upon trying, I now realise that I need to rework the poms so that the bom can satisfy the central release requirements, so its going to be Monday.

gemmellr:
I've pushed an updated+squashed version of the PR now, using the updated extension release with bom.

gemmellr:
I've raised the related PR raised for updating the JMS guide: https://github.com/quarkusio/quarkus/pull/7538

gsmet:
@aloubyansky I let you approve that one? Or should I have a closer look?

PR #31, opened 2020-03-17
add kogito test module

mariofusco:


mariofusco:
@gsmet @lburgazzoli The failure seems totally unrelated with this pull request. It's failing in the module `quarkus-universe-integration-tests-camel-consul` with `java.lang.ClassNotFoundException: io.quarkus.deployment.devmode.HotReplacementContext`. I believe that the original test module hosted in the camel codebase is changed and now it is making this to fail. Any feedback?

lburgazzoli:
As far as I remember HotReplacementContext has been moved to io.quarkus.dev.spi in 1.3.0 but the camel version which is set in the platform is based on 1.2.1.

We'll update the camel version as soon as the new release is available on maven central

gsmet:
@lburgazzoli I updated Camel Quarkus here: https://github.com/quarkusio/quarkus-platform/pull/32

Will wait for CI and merge.

kogito-integration-tests/kogito-quarkus-integration-test/javax.inject.Named
```diff
@@ -0,0 +1,2 @@
+io.quarkus.it.kogito.jbpm.OrderItemsProcess
+io.quarkus.it.kogito.jbpm.OrdersProcess
```
gsmet: This shouldn't be in.

It looks like an issue we had in the past with sisu-inject being a dependency and going berserk. Can you check?

See https://github.com/quarkusio/quarkus/pull/2889 and https://github.com/quarkusio/quarkus/pull/2986

kogito-integration-tests/kogito-quarkus-integration-test/pom.xml
```diff
@@ -0,0 +1,248 @@
...
+                <exclusion>
+                    <groupId>*</groupId>
+                    <artifactId>*</artifactId>
+                </exclusion>
+            </exclusions>
+        </dependency>
+        <dependency>
+            <groupId>org.kie.kogito</groupId>
+            <artifactId>kogito-quarkus-integration-test</artifactId>
+            <version>0.8.0</version>
```
gsmet: Can you make this a property in the parent pom?

kogito-integration-tests/kogito-quarkus-integration-test/pom.xml
```diff
@@ -0,0 +1,248 @@
...
+        <dependency>
+            <groupId>org.kie.kogito</groupId>
+            <artifactId>kogito-dmn</artifactId>
+            <version>0.8.0</version>
+        </dependency>
+        <dependency>
+            <groupId>org.kie.kogito</groupId>
+            <artifactId>kogito-ruleunits</artifactId>
+            <version>0.8.0</version>
+        </dependency>
```
gsmet: I'm a bit surprised you have all those dependencies? Shouldn't they come with the IT dependency? Same for the Quarkus ones below.

kogito-integration-tests/kogito-quarkus-integration-test/javax.inject.Named
mariofusco: @gsmet right, I'll give a look

kogito-integration-tests/kogito-quarkus-integration-test/pom.xml
mariofusco: @gsmet this pom is automatically generated. Yes, I will make it a property, but it won't change much imo.

kogito-integration-tests/kogito-quarkus-integration-test/pom.xml
mariofusco: @gsmet It also surprised me a bit that I had to do this. The problem is that if I don't do that the tests fail with some `ClassNotFoundException` or `NoClassDefFoundError` and I haven't been able to understand why. Any clue? Something related with Quarkus classloader?

mariofusco:
Any news on this?

kogito-integration-tests/kogito-quarkus-integration-test/pom.xml
```diff
@@ -0,0 +1,257 @@
...
+        <groupId>io.quarkus</groupId>
+        <artifactId>quarkus-universe-kogito-integration-tests-parent</artifactId>
+        <version>999-SNAPSHOT</version>
+        <relativePath>../pom.xml</relativePath>
+    </parent>
+
+    <artifactId>kogito-quarkus-integration-test</artifactId>
+
+    <properties>
+        <kogito.version>0.8.0</kogito.version>
```
gsmet: I was thinking of `quarkus-universe-parent` together with the Camel Quarkus version.

kogito-integration-tests/kogito-quarkus-integration-test/pom.xml
gsmet: Are they test dependencies of the IT? If so, they won't be transitive. I think it would make sense to not make them test dependencies of the IT.

kogito-integration-tests/kogito-quarkus-integration-test/pom.xml
gsmet: Hmmm, and indeed you added it in the main parent pom. Is there a reason it's duplicated here?

mariofusco: @gsmet The problem here is that I have to add those dependencies explicitly (together with their version) because otherwise the quarkus bom 1.3.0 will bring in kogito 0.7.1 and this will break the tests. This problem will be solved with quarkus 1.4 which won't contain kogito anymore.

mariofusco:
replaced by https://github.com/quarkusio/quarkus-platform/pull/34

PR #35, opened 2020-03-24
Tentative list of supported extensions for RHBQ

tqvarnst:
This PR contains a tentative list of labels for Red Hat build of Quarkus. This will likely require change as we get closer to GA, but we can use this to test and verify code.quarkus.redhat.com


maxandersen:
thanks @tqvarnst I merged it with my PR #36 as I can see I gave you the wrong attribute names so I had do a global search and replace ;)

PR #36, opened 2020-03-24
initial multi overrides

maxandersen:
should not be merged before  https://github.com/quarkusio/quarkus/pull/8101 is ready.

### DIFF
--- a/bom/runtime/src/main/resources/extensions-support-overrides.json
+++ b/bom/runtime/src/main/resources/extensions-support-overrides.json
@@ -1,0 +1,404 @@
+{
+  "extensions" :[
+    {
+      "group-id": "io.quarkus",
+      "artifact-id":"quarkus-agroal",
+      "metadata" : { 
+       "redhat-support" : [
+         "supported" ]
+       }
+    },
+    {
+      "group-id": "io.quarkus",
+      "artifact-id":"quarkus-config-yaml",
+      "metadata" : { 
+       "redhat-support" : [
+         "supported" ]
+       }
+    },
+    {
+      "group-id": "io.quarkus",
+      "artifact-id":"quarkus-hibernate-orm",
+      "metadata" : { 
+       "redhat-support" : [
+         "supported" ]
+       }
+    },
+    {
+      "group-id": "io.quarkus",
+      "artifact-id":"quarkus-hibernate-orm-panache",
+      "metadata" : { 
+       "redhat-support" : [
+         "supported" ]
+       }
+    },
+    {
+      "group-id": "io.quarkus",
+      "artifact-id":"quarkus-hibernate-validator",
+      "metadata" : { 
+       "redhat-support" : [
+         "supported" ]
+       }
+    },
+    {
+      "group-id": "io.quarkus",
+      "artifact-id":"quarkus-jackson",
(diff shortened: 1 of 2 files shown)


maxandersen:
waiting - found bug in the metatadata

maxandersen:
data fixed.

bom/runtime/src/main/resources/extensions-support-overrides.json
```diff
@@ -0,0 +1,454 @@
+{
+  "extensions" :[
+    {
+      "group-id": "io.quarkus",
+      "artifact-id":"quarkus-agroal",
+      "metadata" : { 
+       "supported-by" : "redhat",
```
emmanuelbernard: "redhat", not "Red Hat" OMG ;)

bom/runtime/src/main/resources/extensions-support-overrides.json
```diff
@@ -0,0 +1,454 @@
...
+       "redhat-support" : [
```
emmanuelbernard: I like this notion of prefix `redhat-` or `redhat.` to avoid namespace collision for vendor specific settings

bom/runtime/src/main/resources/extensions-support-overrides.json
maxandersen: it’s a key :)

bom/runtime/src/main/resources/extensions-support-overrides.json
maxandersen: yes, that’s the idea

bom/runtime/src/main/resources/extensions-support-overrides.json
```diff
@@ -0,0 +1,454 @@
...
+         "supported" ]
+       }
+    },
+    {
+      "group-id": "io.quarkus",
+      "artifact-id":"quarkus-smallrye-context-propagation",
+      "metadata" : { 
+       "supported-by" : "redhat",
```
michalszynkiewicz: I think sth like:
```
"support" : [
   {
      "redhat": [ "supported" ]
   },
]
```
or even:
```
"support" : [
   {
      "redhat": { "supported":true, "deprecated": true }
   }, ...
]
```
would be a bit prettier (to not have two `redhat`s in the metadata)

maxandersen: @michalszynkiewicz I like your first proposal; second one about making boolean isn't really adding value besides requiring more characters ;)

After talking with @ia3andy we simplified it and suggest we go with:

```
"support" : 
   {
      "redhat": [ "supported" ]
   },
```

basically a map rather than a list.

emmanuelbernard: Not sure how the list of support elements can nicely be overridden in an additive way with our current overriding mechanism, so we would override everything.

maxandersen: yeah so back and forth a few times we are ending up with 

```
"redhat-support" : [ "supported", ... ]
```

no supported-by and tools/code.quarkus.io will have a optoin to state which key to use for status.

PR #40, opened 2020-04-06
Add openshift extension next to the lubernetes one in code.quarkus.io

emmanuelbernard:
I found it odd UX wise that Kubernetes was pinned up and that the OpenShift one ended up being last. This fix put them together. CC @maxandersen


maxandersen:
makes sense!

emmanuelbernard:
merging?

PR #44, opened 2020-04-09
Reenable Infinispan tests and disable MongoDB tests

gsmet:


src/rpkg-templates/run-tests-module-pom.xml
```diff
@@ -10,7 +10,7 @@
 
     <artifactId>[=runTestsModule.artifactId]</artifactId>
 
-[#if runTestsModule.artifactId = 'quarkus-universe-integration-tests-camel-infinispan']
+[#if runTestsModule.artifactId = 'quarkus-universe-integration-tests-camel-mongodb']
```
ppalaga, line 13: Hm... I am afraid, `skipTests=true` will not be enough for the mongo tests. We are actually disabling the mongo test on the Camel quarkus side https://github.com/apache/camel-quarkus/pull/1063/files#diff-32e5ef896402deb1c95ac6b572035786R94 We need to take care on the Camel Quarkus side that the mongo test won't be exported to the testJars file. Thanks for pinging me, I need to do that. Once I do it, there is no need for any special handling here (the mongo module will disappear upon re-running camel-regen).

gsmet: Well, it works. And the idea is to reenable it later.

gsmet: But I don't really care for the solution. I just want to be able to release the platform.

ppalaga:
Superceded by https://github.com/quarkusio/quarkus-platform/pull/49. Please close this one.

gsmet:
Closing as way out of date.

PR #45, opened 2020-04-09
Bump Kogito to 0.9.0

evacchi:
@mariofusco can you double check this is all that's required?

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -56,7 +56,7 @@
 
         <quarkus-qpid-jms.version>0.13.0</quarkus-qpid-jms.version>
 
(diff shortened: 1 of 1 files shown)


gsmet:
That's not sufficient, you also need to regen the ITs `mvn validate -Pregen-kogito -N` (I added a comment so that you don't miss it next time in #46 ).

But better rebase on master once https://github.com/quarkusio/quarkus-platform/pull/46 is in as I moved the Kogito tests to another place and it's going to conflict.

evacchi:
waiting on the release of 0.9.1 (this week)

PR #49, opened 2020-04-14
Reenable Camel Infinispan tests and remove the special handling for camel-core-main

ppalaga:


gsmet:
@ppalaga as done in my other PR, we need to disable the MongoDB tests. I would prefer it done in this PR too.

ppalaga:
> we need to disable the MongoDB tests

We made some changes in the Camel Quarkus side so that the mongo test is not exported into test list. It is this file for CQ 1.0.0-M6 https://repo1.maven.org/maven2/org/apache/camel/quarkus/camel-quarkus-test-list/1.0.0-M6/camel-quarkus-test-list-1.0.0-M6.xml Mongo is still listed there, but if you check the file for CQ master, it is not there anymore and thus the module will be removed by `mvn -N clean validate -Pregen-camel`. Is that solution not suffcient for some reason?

PR #55, opened 2020-04-21
Kogito 0.10.0 upgrade

gsmet:


integration-tests/kogito/kogito-quarkus-integration-test/pom.xml
```diff
@@ -231,7 +231,7 @@
                                     <goal>native-image</goal>
                                 </goals>
                                 <configuration>
-                                    <appArtifact>org.kie.kogito:kogito-quarkus-integration-test:0.8.0</appArtifact>
+                                    <appArtifact>org.kie.kogito:kogito-quarkus-integration-test:0.10.0</appArtifact>
```
aloubyansky, line 234: Shouldn't it be `${kogito-quarkus.version}` instead of the specific value? The same everywhere else.

PR #56, opened 2020-04-23
1.4 mess

aloubyansky:


geoand:
Seems like it passed... @gsmet did you see this?

aloubyansky:
I'm just messing here. This is narrowed down version of the TS. This test used to pass before.

aloubyansky:
It downloaded a lot of 1.3.X deps though.

PR #58, opened 2020-04-23
Migrate to GitHub actions

geoand:


geoand:
Kogito fails which according to @gsmet is to be expected for now

maxandersen:
@geoand anything blocking this for going in ?

geoand:
I totally forgot about this one...

Let me rebase

geoand:
Force pushed an update

gsmet:
So I would like us to be consistent and use GH Actions everywhere. My only issue is that when doing a release, I don't want to be blocked by the queue being full of Quarkus Core CI builds.

We don't have a way to keep a few jobs free for important tasks with GH Actions so I wonder if we should wait for having a proper solution before doing the switch.

gsmet:
@geoand I just refreshed and improved this PR as we have more and more issues with Azure. Even if it might be a bit of pain to cancel already planned jobs to get this tested when releasing, it will still be better than having our builds failing.

I'll see how the CI build goes.

geoand:
Cool, I had forgot about this completely :)

PR #69, opened 2020-05-14
Avoid Connection timed out (Read failed) on Azure by setting maven.wa…

ppalaga:
…gon.httpconnectionManager.ttlSeconds

@geoand we were struggling with this kind of issues in Camel Quarkus for some time. 

There is some background info in https://issues.apache.org/jira/browse/WAGON-486 and https://issues.apache.org/jira/browse/WAGON-545

I hope azure pipelines is using a Maven version new enough to pick `./mvn/maven.config`

### DIFF
--- a/.mvn/maven.config
+++ b/.mvn/maven.config
@@ -1,0 +1,3 @@
+-Dmaven.wagon.httpconnectionManager.ttlSeconds=120
+-Dmaven.wagon.http.retryHandler.requestSentEnabled=true
+-Dmaven.wagon.http.retryHandler.count=10


geoand:
That's nice! 

Do we perhaps need to do anything else to make https://github.com/quarkusio/quarkus-platform/blob/master/.github/quarkus-ecosystem-test#L5 pick up these settings?

ppalaga:
> That's nice!
> 
> Do we perhaps need to do anything else to make https://github.com/quarkusio/quarkus-platform/blob/master/.github/quarkus-ecosystem-test#L5 pick up these settings?

No, if you use maven 3.3.1+ https://maven.apache.org/docs/3.3.1/release-notes.html#jvm-and-command-line-options

famod:
@ppalaga Any reason why you didn't use `.mvn/jvm.config` instead? My understanding from the Maven docs is that `maven.config` should be used for `mvn` arguments like `--fail-at-end` etc.
Reason I'm asking is that I want to set `-Dkotlin.environment.keepalive=true` in the main repo to work around https://youtrack.jetbrains.com/issue/KT-45007

ppalaga:
`-D` options might work in both files hard to say whether it makes any difference.

famod:
Ok, thanks for your feedback.
FWIW, it seems `jvm.config` is added to `MAVEN_OPTS` somewhere in the scripts whereas `maven.config` is read in some CLI class in maven-embedder. So I guess things like locale etc. that have to be set early on to take effect _have_ to be in `jvm.config`.

PR #70, opened 2020-05-14
 Enable batch mode to avoid a too verbose output

geoand:
Same as https://github.com/quarkusio/quarkus-ecosystem-ci/pull/14

### DIFF
--- a/.github/quarkus-ecosystem-test
+++ b/.github/quarkus-ecosystem-test
@@ -2,4 +2,4 @@
 set -e
 
 # just run all the tests
-mvn --settings .github/quarkus-ecosystem-maven-settings.xml clean install -Dquarkus.version=${QUARKUS_VERSION}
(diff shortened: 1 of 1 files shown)


ppalaga:
`-ntp` is another good candidate https://maven.apache.org/docs/3.6.1/release-notes.html#user-visible-changes

geoand:
Ah, nice!

geoand:
It seems like `-B` is a superset of `-ntp`, no?

ppalaga:
> It seems like `-B` is a superset of `-ntp`, no?

I think `-B` there is still a lot of `Download[ing|ed] <url>` lines in the log that are not there with `-ntp`

geoand:
I'll check tomorrow morning after the job has run.

If I see those entries, I'll add `-ntp` as well

PR #80, opened 2020-06-24
Add cassandra-client to quarkus-platform

adutra:
This is a request to add the [Cassandra Quarkus extension](https://github.com/datastax/cassandra-quarkus) to the platform.

See also https://github.com/quarkusio/quarkus/issues/7876.

\cc @emmanuelbernard @tomekl007 @absurdfarce


emmanuelbernard:
info for @gsmet and @aloubyansky 
We agreed to add Cassandra as experimental. The alpha feels scary but they have done testings with a few users already so it's just a conservative alpha label.

integration-tests/cassandra-client/pom.xml
```diff
@@ -0,0 +1,162 @@
...
+                <id>native-tests</id>
+                <goals>
+                  <goal>integration-test</goal>
+                  <goal>verify</goal>
+                </goals>
+                <configuration>
+                  <groups>native</groups>
+                  <excludes>
+                    <!-- this test is broken in 1.0.0-alpha1 -->
+                    <exclude>**/CassandraMetricsNativeIT.java</exclude>
```
adutra, line 148: This test will be fixed for the next alpha, see https://github.com/datastax/cassandra-quarkus/issues/96.

gsmet:
I triggered a re-run on CI. CI failed in Camel Quarkus Debezium.

integration-tests/cassandra-client/pom.xml
```diff
@@ -0,0 +1,162 @@
...
+      <version>${quarkus-cassandra-client.version}</version>
+      <scope>test</scope>
+    </dependency>
+    <!-- Need to force version 4.2, 4.3 has binary incompatibility issues -->
+    <dependency>
+      <groupId>io.rest-assured</groupId>
+      <artifactId>rest-assured</artifactId>
+      <version>4.2.0</version>
+      <scope>test</scope>
+    </dependency>
```
aloubyansky, line 43: What kind of incompatibility issues are there?
This is alarming since here you are walking around the platform BOM to get the TS to pass. But when a user creates a project using the platform BOM, the platform version of the dependency will be used instead.

gsmet: I think it's probably just that their ITs are still written with Rest Assured 4.2. I wouldn't expect it to be used at runtime. Am I right?

aloubyansky: Right, it's most probably the case. Still will be good clarify. This is the default for Quarkus apps.

adutra: You are both correct, it's certainly us using something that was removed in Rest Assured 4.3. More specifically the error is:

```
java.lang.AbstractMethodError: io.restassured.internal.ResponseParserRegistrar.getProperty(Ljava/lang/String;)Ljava/lang/Object;
	at org.codehaus.groovy.runtime.callsite.PogoGetPropertySite.getProperty(PogoGetPropertySite.java:49)
	at org.codehaus.groovy.runtime.callsite.AbstractCallSite.callGroovyObjectGetProperty(AbstractCallSite.java:309)
	at io.restassured.internal.ResponseParserRegistrar.<init>(ResponseParserRegistrar.groovy:44)
	at io.restassured.RestAssured.createTestSpecification(RestAssured.java:1448)
	at io.restassured.RestAssured.given(RestAssured.java:653)
...
```

However it wasn't easy to find the incriminated method call because there is some groovy involved, and as you pointed out, this is a test dependency only, and we don't impose it on users of our extension, so eventually we gave up on trying to fix this for now.

Maybe @absurdfarce has more details.

aloubyansky: Ok, that's good enough. Thanks.

gsmet:
There's one small issue though. We need a `<description>` in the extension runtime module here: https://github.com/datastax/cassandra-quarkus/blob/master/runtime/pom.xml#L27 . This is what is used here in front of the extension name: https://code.quarkus.io/ . Ideally it would start with a verb: "Connect to ..." or whatever.

Also do you plan to work on a guide soon? If not I wouldn't add the guide here until it's done: https://github.com/datastax/cassandra-quarkus/blob/master/runtime/src/main/resources/META-INF/quarkus-extension.yaml#L9 .

I won't block the inclusion for that but if you can fix these issues before the 1.6 Platform release (so before July 7th), that would be great.

Thanks!

integration-tests/cassandra-client/pom.xml
absurdfarce: FTR there's some additional detail on the issues around this test in https://github.com/datastax/cassandra-quarkus/issues/28

gsmet:
@adutra could you ack you saw my comment above when you see it? Thanks!

adutra:
@gsmet 

> There's one small issue though. We need a `<description>` in the extension runtime module here: https://github.com/datastax/cassandra-quarkus/blob/master/runtime/pom.xml#L27 . This is what is used here in front of the extension name: https://code.quarkus.io/ . Ideally it would start with a verb: "Connect to ..." or whatever.

OK, added.

> Also do you plan to work on a guide soon? If not I wouldn't add the guide here until it's done: https://github.com/datastax/cassandra-quarkus/blob/master/runtime/src/main/resources/META-INF/quarkus-extension.yaml#L9 .

We already have a guide, only the sources are in our repo, not yours. In the link below you will find the asciidoc source and the generated html and pdf:

https://github.com/datastax/cassandra-quarkus/releases/tag/1.0.0-alpha1

Let me know if you need us to change anything in the guide.

Thanks!

gsmet:
OK, so if you can push a new release before July 7th with the description fix, that would be awesome. Otherwise it will look a bit ugly on code.quarkus.io.

As for the guide, if you want, we can push it to the main repo. For now, that's the only way to host it on our website. We need to figure out how we can distribute that but we're not there yet.

It would have to be adapted a bit (mostly to have the same header as the other guides and the name would need to be cassandra.doc to be in line with the URL you gave).

Another option would be that you host your guide yourself and you point to your own URL in the YAML descriptor. That would work too.

Either way, we don't want the guide pointing to a 404 so choose the best option for you and let me know :).

adutra:
> OK, so if you can push a new release before July 7th with the description fix, that would be awesome. Otherwise it will look a bit ugly on code.quarkus.io.

We will publish another alpha next week, and the fix will be in there.

> As for the guide, if you want, we can push it to the main repo. For now, that's the only way to host it on our website. We need to figure out how we can distribute that but we're not there yet.

That was our expectation, so yes, if you can push it to the main repo that would be great. (Preferably next week, after our next alpha release.) We commit to maintaining that file in sync with our releases going forward by opening PRs whenever necessary.
 
> It would have to be adapted a bit (mostly to have the same header as the other guides and the name would need to be cassandra.doc to be in line with the URL you gave).

I will rename the file as you suggested to `cassandra.adoc`. However what modification to the contents do you wish to have? We tried to make our doc very similar to the [MongoDB one](https://raw.githubusercontent.com/quarkusio/quarkus/9eb5164f24a2c3bc6b4015bc4e399a7ef955704c/docs/src/main/asciidoc/mongodb.adoc), including headers.

PR #82, opened 2020-06-29
Upgrade to Quarkus 1.6.0.CR1 and make sure the old versions are excluded from the BOM and extensions.json

gsmet:
@aloubyansky are there other things I need to exclude? AFAICS the Optaplanner ones are the only ones in 1.5.0.Final.

### DIFF
--- a/bom/deployment/pom.xml
+++ b/bom/deployment/pom.xml
@@ -176,6 +176,102 @@
                             </regexFlags>
                         </configuration>
                     </execution>
+                    <execution>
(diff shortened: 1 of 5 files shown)


aloubyansky:
Here is the complete report https://filebin.net/xe8objoo2r1giyv8/bom-report.html?t=r1k5bao9

gsmet:
I'm not sure I understand fully what's going on in this file.

gsmet:
@aloubyansky I added more commits. With the core PR and this one, it should improve the situation a lot.

PR #81, opened 2020-06-28
Work around testcontainers leaking in from various BOMs

famod:
Fixes https://github.com/quarkusio/quarkus/issues/9968 (workaround) until the mentioned upstream BOMs are adjusted.

See also: https://github.com/quarkusio/quarkus/issues/9968#issuecomment-650835627

PR #83, opened 2020-06-29
Add some enforcer rules

gsmet:
**Obviously, don't merge because CI will fail.**

Add the dependency convergence rule and also the banned dependencies we
have in Quarkus.

The banned dependencies might not be all relevant but it's a start to
figure out what's going on.

/cc @ppalaga @aloubyansky this doesn't work very well :). I have a lot of errors in quarkus-universe-integration-tests-camel-rpkgtests . I'm not sure the dependency convergence rule makes sense because you probably won't use all the Camel components at once but IMHO that would be good to fix that.

Also the banned dependencies issues should be fixed IMHO. These dependencies were banned for a good reason.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -126,6 +126,111 @@
                     <artifactId>rpkgtests-maven-plugin</artifactId>
                     <version>${rpkgtests-maven-plugin.version}</version>
                 </plugin>
+                <plugin>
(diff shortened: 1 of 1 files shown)


famod:
Impressive blacklist. ;-)

What about testcontainers?
See also: https://github.com/quarkusio/quarkus-platform/pull/81

PR #88, opened 2020-07-02
Manage debezium and testcontainers on the top level to solve conflicts

ppalaga:
among participants

@gsmet We'd need something like this to make our debezium tests work.

### DIFF
--- a/bom/runtime/pom.xml
+++ b/bom/runtime/pom.xml
@@ -28,6 +28,51 @@
                 <scope>import</scope>
             </dependency>
 
+            <!-- Resolve conflicts among platform participants -->
+            <dependency>
+                <groupId>io.debezium</groupId>
+                <artifactId>debezium-api</artifactId>
+                <version>${debezium.version}</version>
+            </dependency>
+            <dependency>
+                <groupId>io.debezium</groupId>
+                <artifactId>debezium-connector-mysql</artifactId>
+                <version>${debezium.version}</version>
+            </dependency>
+            <dependency>
+                <groupId>io.debezium</groupId>
+                <artifactId>debezium-connector-postgres</artifactId>
+                <version>${debezium.version}</version>
+            </dependency>
+            <dependency>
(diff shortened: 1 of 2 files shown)


ppalaga:
The bottom line is that IMO testcontainers and other testing-only related dependencies should be banned from the platform BOM and it should be individual test participants' responsibility to add/manage them individually. In such a case, we should still check that those participant's test BOM do not override anything from the Platform BOM.

bom/runtime/pom.xml
```diff
@@ -28,6 +28,51 @@
...
+                <artifactId>debezium-embedded</artifactId>
+                <version>${debezium.version}</version>
+            </dependency>
+            <dependency>
+                <groupId>org.testcontainers</groupId>
+                <artifactId>testcontainers-bom</artifactId>
+                <type>pom</type>
+                <scope>import</scope>
+                <version>1.14.3</version>
+            </dependency>
```
gsmet, line 73: I'm not entirely sure this is a good idea to have it here given we decided to remove it from the Quarkus core BOM.

@aloubyansky what's your take on that one?

gsmet: BTW, additional question: is it a test only dependency?

aloubyansky: Whether it's only a test dep is the key question. Are they all necessary in the user-facing BOM @ppalaga ?

famod: Just to make sure: Have you guys seen this: https://github.com/quarkusio/quarkus/issues/9968#issuecomment-650835627
and the resulting PR #81 here in quarkus-platform?

gsmet: @famod thanks, I missed it and your analysis is indeed very useful. I'll try to come up with a contingency plan for 1.6 and a better plan for 1.7.

gsmet:
Closing in favor of https://github.com/quarkusio/quarkus-platform/pull/92 .

PR #93, opened 2020-07-07
Upgrade cassandra-client to 1.0.0-alpha3

adutra:
Probably a bit too late for Quarkus 1.6.0 but :-)

This new alpha version simply follows received feedback and turns a few dependencies from optional to mandatory (smallrye-health and smallrye-metrics).

@gsmet I'm not seeing the cassandra.adoc file here:

https://github.com/quarkusio/quarkusio.github.io/tree/develop/_guides

Am I looking at the right place? Thanks!


gsmet:
Thanks. Yeah, I need to address the Cassandra guide. Will do that tomorrow when I publish the guides for 1.6.0.Final.

PR #94, opened 2020-07-08
Use 1.0.0-RC5 version of quarkus-hazelcast-client

pivovarit:


gsmet:
Thanks, let me backport it to 1.6 so that's it's part of 1.6.1.Final.

pivovarit:
Would be great. Quite a serious regression sneaked into the previous one.

PR #120, opened 2020-09-10
Bump to Kogito 0.15.0

evacchi:
I am actually bumping to 1.0.0-SNAPSHOT (master nightly). We should release 0.15.0 on Friday

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -65,7 +65,7 @@
 
              mvn validate -Pregen-kogito -N
         -->
-        <kogito-quarkus.version>0.13.1</kogito-quarkus.version>
+        <kogito-quarkus.version>0.15.0</kogito-quarkus.version>
         <optaplanner-quarkus.version>7.43.0.Final</optaplanner-quarkus.version>
         <testcontainers.version>1.14.3</testcontainers.version>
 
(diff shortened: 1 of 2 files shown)


evacchi:
alright, so I suppose snapshots won't work :sweat_smile:

pom.xml
```diff
@@ -82,6 +82,19 @@
         <module>integration-tests</module>
     </modules>
 
+    <repositories>
+        <repository>
+            <id>jboss-public-repository-group</id>
+            <name>JBoss Public Repository</name>
+            <snapshots>
+              <enabled>false</enabled>
```
gsmet, line 90: Maybe you should enable the snapshots if you want them to work :].

evacchi: 😅😅😅😅

evacchi:
ok kogito tests ran successfully. I am not sure what's the deal with optaplanner though -- but it says something about the snapshot package, so maybe that's just about the snapshot being used there.

gsmet:
I think it probably got this artifact from the JBoss repository and you forcefully disabled the releases there.

evacchi:
I disabled it because the previous run took >60 minutes

PR #125, opened 2020-09-16
Generate the platform BOM and use Maven invoker to launch the IT of every extension

aloubyansky:
This PR integrates Quarkus Platform BOM Generator [1] and re-structures the repository to actually use the generated BOM in place of the original platform BOM, i.e. its original pom.xml. That is achieved by separating/disconnecting the extension IT projects from the main Quarkus platform project and using the maven-invoker-plugin to launch the tests.

FYI, I had to branches. In one I combined all the extension ITs under a single root and targeted that common root with the maven-invoker-plugin. But it was difficult to track failures in that setup.
In this PR I separated each extension IT into their own project and called maven-invoker-plugin on them individually.

[1] https://github.com/quarkusio/quarkus-platform-bom-generator


aloubyansky:
FYI @gsmet 
FYI @ppalaga we discussed the possibility of generating the BOM, so this is the attempt to introduce it.

aloubyansky:
Updated the readme describing how it's organized.

ppalaga:
Looks OK visually. Trying locally now...

gsmet:
I haven't looked at the patch but quick question, did you also adjust the `regen` part used to refresh the ITs?

integration-tests/pom.xml
```diff
@@ -59,14 +58,49 @@
...
         <pluginManagement>
             <plugins>
+                <plugin>
+                    <artifactId>maven-invoker-plugin</artifactId>
+                    <version>3.2.1</version>
+                    <configuration>
+                        <projectsDirectory>invoked</projectsDirectory>
+                        <streamLogs>true</streamLogs>
+                        <suppressSummaries>true</suppressSummaries>
+                        <goals>clean install -e</goals>
```
ppalaga, line 86:
```suggestion
                        <goals>clean verify -e</goals>
```
There is no reason to install, is there?

ppalaga: Hm... there seem to be no way to perform the native tests. I think we should define a top level `<invoker.goals>` property so that it is overridable on the CLI - something like `mvn clean verify "-Dinvoker.goals=clean verify -Dnative"`. WDYT?

ppalaga:
> did you also adjust the `regen` part used to refresh the ITs?

It does not seem to be the case. I can fix it (for Camel at least).

ppalaga:
Note to myself: running a single test is possible via `mvn clean verify -Dinvoker.test=root/camel-activemq`

gsmet:
It needs to be fixed for everyone before we merge this.

gsmet:
> Note to myself: running a single test is possible via mvn clean verify -Dinvoker.test=root/camel-activemq

Maybe we could add that to the README?

integration-tests/pom.xml
aloubyansky: Ah, right, i missed the native args. About verify, I'm not sure it'll work if you clean your local repo.

ppalaga: I have sent https://github.com/aloubyansky/quarkus-platform/pull/1 to your fork, @aloubyansky

ppalaga:
> > did you also adjust the `regen` part used to refresh the ITs?
> 
> It does not seem to be the case. I can fix it (for Camel at least).

> did you also adjust the `regen` part used to refresh the ITs?

Fixed for Camel https://github.com/aloubyansky/quarkus-platform/pull/2

integration-tests/pom.xml
ppalaga: > About verify, I'm not sure it'll work if you clean your local repo.

I can try that tomorrow.

ppalaga:
One more idea: I'd find it handy to have the `bom/target/boms/*/pom.xml` tracked in git (but the version should not be a part of the path). A history of those files would make it easier to figure out which upgrade caused some specific problem.

integration-tests/pom.xml
aloubyansky: So if you want to be able to `mvn clean verify -Dinvoker.test=root/camel-activemq` you need `clean install` as invoker goals. Otherwise, it won't find the rpkgtests artifact.

aloubyansky:
> > Note to myself: running a single test is possible via mvn clean verify -Dinvoker.test=root/camel-activemq
> 
> Maybe we could add that to the README?

I've added a paragraph about running tests https://github.com/aloubyansky/quarkus-platform/tree/invoker-per-extension#running-tests

aloubyansky:
> > > did you also adjust the `regen` part used to refresh the ITs?
> > 
> > 
> > It does not seem to be the case. I can fix it (for Camel at least).
> 
> > did you also adjust the `regen` part used to refresh the ITs?
> 
> Fixed for Camel [aloubyansky#2](https://github.com/aloubyansky/quarkus-platform/pull/2)

Merged. Thanks!

There was another regen profile for kogito, which I've adjusted. I have not introduced equivalent profiles for other extensions.

aloubyansky:
> One more idea: I'd find it handy to have the `bom/target/boms/*/pom.xml` tracked in git (but the version should not be a part of the path). A history of those files would make it easier to figure out which upgrade caused some specific problem.

That's an interesting idea. I'd probably keep it as a separate issue though. Given that they are in the target dir, they could easily (intentionally) be removed by simply running `clean`. Worth investigating though.

integration-tests/pom.xml
ppalaga: Yeah, I have noticed that. I wonder whether it would be possible to move the rpkg module out of the invoker dir and execute it within the main build.

aloubyansky: We'd need all the extension test jars configured in the main build then. Not sure, tbh. Needs more evaluation.

aloubyansky:
So, does this look good for the first iteration?

ppalaga:
I see no camel tests executed on the CI - is that expected?

ppalaga:
Camel tests are passing for me locally.

aloubyansky:
> I see no camel tests executed on the CI - is that expected?

Not sure, when I run `mvn clean install` i see them running.

aloubyansky:
the CI is running `clean package` though.

aloubyansky:
ah, `clean package` is not triggering the tests, good catch :)

aloubyansky:
i think it should be `verify -pl integration-tests`

aloubyansky:
No, that didn't work. Simply `verify` works. But we don't need to regenerate the BOM.

aloubyansky:
`mvn verify -pl '!bom'` appears to work.

aloubyansky:
Now I actually see them running in CI.

integration-tests/pom.xml
ppalaga: Thinking more about that, it could certainly be set up, but it is not a priority ATM. I'll do it in the future in case the current setup becomes too limiting for me.

ppalaga:
Yeah, I can see them running too. Good to merge from my POV. Thanks a lot @aloubyansky !

ppalaga (approved):

gsmet (approved):
@aloubyansky I ran all the quickstarts tests with this and it works flawlessly.

Should I merge it?

aloubyansky:
Unless anybody has any other remarks, I say yes.

gsmet:
Thanks!

PR #129, opened 2020-10-05
Added RHBQ 1.7 extensions

tqvarnst:
This is the expect list of new supported and Tech-preview extensions for RHBQ 1.7


ia3andy (approved):
LGTM

rsvoboda:
@tqvarnst @ia3andy @gsmet this went to 1.7, is it in master too?

Similar case to https://github.com/quarkusio/quarkus-platform/pull/132

gsmet:
My inclination would be to include this in master too as it will be the bottom line for next product release. @tqvarnst WDYT?

PR #132, opened 2020-10-07
Minor fixes to the overide for RHBQ 1.7

tqvarnst:
Added missing override for `quarkus-resteasy-qute` and fixed artifact-id for `quarkus-reactive-mysql-client`.


gsmet (approved):

gsmet:
hey @tqvarnst @ia3andy 

So this got merged to 1.7 but we don't have the same changes in master. Is it OK? We won't need these overrides for the next product version?

Thanks for confirming.

PR #154, opened 2020-11-24
Exclude org.testcontainers dependencies from the BOM

aloubyansky:
This PR removes the testcontainers dependencies from the BOM.
/cc @evacchi @mariofusco

### DIFF
--- a/bom/pom.xml
+++ b/bom/pom.xml
@@ -200,20 +200,20 @@
                         </goals>
                     </execution>
                 </executions>
-                <!--
-                Templates for enforcing specific versions and excluding dependencies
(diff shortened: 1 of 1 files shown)


evacchi:
@aloubyansky that should be fixed in my pr for kogito 1.0

gsmet:
@aloubyansky I think this should be merged anyway, right?

aloubyansky:
> @aloubyansky I think this should be merged anyway, right?

It could be, yes. We probably don't want this coming from any where.

evacchi:
No problem, just FYI

evacchi (approved):

PR #159, opened 2020-11-27
Upgrade to the platform BOM generator 0.0.4

aloubyansky:
This upgrade fixes an issue that result in two quarkus-bom platform descriptor artifacts of different versions in the generated BOM.
Apparently, Maven request to resolve an artifact descriptor may return a result for an artifact that does not exist. Now the generator will fail in case that happens.
There was also another issue in the generator related to identifying which kind of quarkus core BOM to subtract (quarkus-bom or quarkus-bom-deployment - there are still extensions using quarkus versions in which the deployment BOM was present).

### DIFF
--- a/bom/pom.xml
+++ b/bom/pom.xml
@@ -191,7 +191,7 @@
             <plugin>
                 <groupId>io.quarkus</groupId>
                 <artifactId>quarkus-platform-bom-maven-plugin</artifactId>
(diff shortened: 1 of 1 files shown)


gsmet:
I confirm I was able to build code.quarkus.io with this patch applied. Thanks!

PR #166, opened 2020-12-16
Update version in pom instead of passing a property

gsmet:
With the current setup, the version is not propagated to the invoked
tests so we end up using the version in the pom.

/cc @aloubyansky

### DIFF
--- a/.github/quarkus-ecosystem-test
+++ b/.github/quarkus-ecosystem-test
@@ -1,5 +1,8 @@
 #!/usr/bin/env bash
 set -e
 
+# update the version
+mvn versions:set-property -pl . -Dproperty=quarkus.version -DnewVersion=999-SNAPSHOT
+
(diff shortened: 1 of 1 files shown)


.github/quarkus-ecosystem-test
```diff
@@ -1,5 +1,8 @@
 #!/usr/bin/env bash
 set -e
 
+# update the version
+mvn versions:set-property -Dproperty=quarkus.version -DnewVersion=999-SNAPSHOT
```
aloubyansky, line 5: BTW, adding `-pl .` will apply it only to the root pom.xml, which is where it is.

gsmet: Done.

aloubyansky (approved):

gsmet:
Let's merge and see how it goes tonight.

PR #168, opened 2020-12-24
Excluding webjars-bootstrap from the BOM

harshmadhani:


bom/pom.xml
```diff
@@ -372,6 +372,22 @@
...
                     </execution>
+                    <execution>
+                        <id>clean-bom-webjars-bootstrap</id>
+                        <phase>process-resources</phase>
+                        <goals>
+                            <goal>replace</goal>
+                        </goals>
+                        <configuration>
+                            <file>${basedir}/.flattened-pom.xml</file>
+                            <token><![CDATA[<dependency>\s*<groupId>org.webjars</groupId>\s*<artifactId>bootstrap</artifactId>\s*<version>3\.1\.0\</version>\s+</dependency>]]></token>
```
gsmet, line 383:
```suggestion
                            <token><![CDATA[<dependency>\s*<groupId>org.webjars</groupId>\s*<artifactId>bootstrap</artifactId>\s*<version>[^<]+</version>\s+</dependency>]]></token>
```
I think we'd better remove all versions to be extra safe.

BTW, are you sure it worked for prod? Because looks like you have an extra `\` at the end of the version.

gsmet (changes requested):

bom/pom.xml
harshmadhani: Yes, we would better remove it for all versions. It did work for prod when we verified the bom. It did not have the bootstrap in the universe bom when built. Do you want to recheck in the next quarkus version if it recoccurs and then merge ? That should be fine too IMHO

gsmet (approved):

bom/pom.xml
gsmet: Works fine, now. Will merge once CI is green, thanks.

PR #183, opened 2021-02-08
Adding supported extensions for 1.11

harshmadhani:
Adding supported extensions for 1.11 cc @tqvarnst


gsmet:
I relaunched CI as one of the job failed because of a CI error.

gsmet:
I will release 1.11.2.Final tomorrow so it would be nice if we could get your approval before 2pm tomorrow @tqvarnst .

gsmet:
OK, well, I will merge this and you can tweak things in the product later on if need be.

harshmadhani:
Thanks @gsmet !

PR #186, opened 2021-02-15
Bump Kogito 1.2 and OptaPlanner 8.2

evacchi:
The 1.2 has been released *while* CR1 got cut. Let's see if it's able to make it without bumping to Quarkus 1.12 explicitly
/cc @mariofusco @cristianonicolai @tiagodolphine @radtriste

### DIFF
--- a/integration-tests/camel/invoked/root/camel-optaplanner/pom.xml
+++ b/integration-tests/camel/invoked/root/camel-optaplanner/pom.xml
@@ -8,6 +8,10 @@
         <relativePath>../pom.xml</relativePath>
     </parent>
 
+    <properties>
+        <skipTests>true</skipTests> <!-- Disabled because of https://github.com/quarkusio/quarkus-platform/pull/186 -->
+    </properties>
+
     <artifactId>quarkus-universe-integration-tests-camel-optaplanner</artifactId>
 
     <dependencies>
--- a/integration-tests/camel/invoked/root/camel-optaplanner/pom.xml
+++ b/integration-tests/camel/invoked/root/camel-optaplanner/pom.xml
@@ -9,7 +9,7 @@
     </parent>
 
     <properties>
-        <skipTests>true</skipTests> <!-- Disabled because of https://github.com/quarkusio/quarkus-platform/pull/186 -->
+        <skipTests>true</skipTests> <!-- Disabled because of https://github.com/apache/camel-quarkus/issues/2253 -->
     </properties>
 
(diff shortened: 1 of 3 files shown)


evacchi:
something about Camel+OptaPlanner is failing @ppalaga @rsynek

ppalaga:
Let me check

ppalaga:
@evacchi so that Camel does not block you, plz feel free to file a new quarkus-platform issue for me and disable the failing test by adding `<skipTests>true</skipTests> <!-- Disabled because of {issue-URL} -->`.

triceo:
As far as I know, this is a bug in OptaPlanner which we'll need to fix.

rsynek:
@ppalaga I can see a CDNF org.optaplanner.core.api.domain.solution.drools.ProblemFactCollectionProperty - which is an annotation from OptaPlanner 7.x, that in OptaPlanner 8 has moved to a different package [1]. Do you have any idea where it comes from in the integration test?

[1] https://www.optaplanner.org/download/upgradeRecipe/upgradeRecipe8.html

integration-tests/camel/invoked/root/camel-optaplanner/pom.xml
```diff
@@ -8,6 +8,10 @@
         <relativePath>../pom.xml</relativePath>
     </parent>
 
+    <properties>
+        <skipTests>true</skipTests> <!-- Disabled because of https://github.com/quarkusio/quarkus-platform/pull/186 -->
```
ppalaga, line 12:
```suggestion
        <skipTests>true</skipTests> <!-- Disabled because of https://github.com/apache/camel-quarkus/issues/2253 -->
```

ppalaga:
> I can see a CDNF org.optaplanner.core.api.domain.solution.drools.ProblemFactCollectionProperty - which is an annotation from OptaPlanner 7.x, that in OptaPlanner 8 has moved to a different package [1]. Do you have any idea where it comes from in the integration test?

It is referenced directly from the test. @zbendhiba might know more.

zbendhiba:
> > I can see a CDNF org.optaplanner.core.api.domain.solution.drools.ProblemFactCollectionProperty - which is an annotation from OptaPlanner 7.x, that in OptaPlanner 8 has moved to a different package [1]. Do you have any idea where it comes from in the integration test?
> 
> It is referenced directly from the test. @zbendhiba might know more.

Yes, I have a workaround for upgrading to 8.x

gsmet (approved):

PR #190, opened 2021-02-19
Fix wrong metadata for 1.11

RotBolt:
cc @harshmadhani  @gsmet

### DIFF
--- a/descriptor/src/main/resources/extensions-support-overrides.json
+++ b/descriptor/src/main/resources/extensions-support-overrides.json
@@ -158,14 +158,6 @@
     },
     {
       "group-id": "io.quarkus",
-      "artifact-id":"quarkus-jdbc-h2",
-      "metadata" : {
-        "redhat-support" : [
-          "dev-support" ]
-      }
-    },
-    {
-      "group-id": "io.quarkus",
       "artifact-id":"quarkus-jsonb",
       "metadata" : { 
        "redhat-support" : [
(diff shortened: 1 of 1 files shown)


harshmadhani (approved):
LGTM

descriptor/src/main/resources/extensions-support-overrides.json
```diff
@@ -156,14 +156,6 @@
...
        }
     },
-    {
-      "group-id": "io.quarkus",
-      "artifact-id":"quarkus-jdbc-h2",
-      "metadata" : {
-        "redhat-support" : [
-          "dev-support" ]
-      }
-    },
```
gsmet, line 166: Why did you remove H2?

Also the original issue is talking about Derby being supported but I don't see it anywhere in the file?

gsmet: @rotbolt ?

tqvarnst: @RotBolt The question here is if we are productizing the h2 extensions (e.g. based on productized artefact). Then it should be marked as Dev supported. I know the tin the sheet we are not listing dev supported, but it depends on if this extension is productized. If it is we have to mark it as Dev supported to make sure that it's not used in production by customers.

RotBolt: Ok I will add back the H2, in the sheet it was not mentioned as Dev support so I removed it.
@gsmet . for the derby issue, derby is actually unsupported as per sheet shared with us
Apologize with late response. I did not get any github notification for this

tqvarnst: I've confirmed with Harsh and @RotBolt that H2 is not productized and therefor should not be marked as dev supported either, so removing them is correct. I believe that the same is true for Derby.

PR #191, opened 2021-02-22
Upgrade to Camel Quarkus 1.7.0

ppalaga:


integration-tests/camel/invoked/root/camel-websocket-jsr356/pom.xml
```diff
@@ -10,11 +10,6 @@
 
     <artifactId>quarkus-universe-integration-tests-camel-websocket-jsr356</artifactId>
 
-    <!-- See: https://github.com/apache/camel-quarkus/issues/2262 -->
-    <properties>
-        <skipTests>true</skipTests>
```
jamesnetherton, line 15: We probably want to keep this since the tests will fail on Quarkus SNAPSHOT builds.

integration-tests/camel/invoked/root/camel-optaplanner/pom.xml
```diff
@@ -8,10 +8,6 @@
         <relativePath>../pom.xml</relativePath>
     </parent>
 
-    <properties>
-        <skipTests>true</skipTests> <!-- Disabled because of https://github.com/apache/camel-quarkus/issues/2253 -->
```
jamesnetherton, line 12: I think this needs to stay as skipped. IIRC the fix for this won't be available until we bump to Camel 3.9.x.

ppalaga: Yeah, I also expected this to fail, but It passed for me locally. Let's see what Platform CI says.

integration-tests/camel/invoked/root/camel-websocket-jsr356/pom.xml
jamesnetherton: Probably because you're not building with `quarkus.version` set to `999-SNAPSHOT`? The breaking changes are in the master branch. Not the 1.12 release.

ppalaga: Oh, I see. Then keeping those tests active is correct as long as we are now building for the release 1.12 of the platform, isn't it?

I have disabled them before I read your message, because it is quite late for any experiments now.

gsmet: No, we need the daily tests with Quarkus master to pass too.

gsmet (approved):

PR #208, opened 2021-03-23
Temporarily disable camel-grpc tests

jamesnetherton:
Relates to latest Quarkus SNAPSHOT build failure:

https://github.com/quarkusio/quarkus/issues/8593#issuecomment-804649551

I'd prefer to only disable these tests on the SNAPSHOT builds, hence the profile hack. Not sure if there's a better way of doing it?


geoand:
Given that it's temporary, I think your fix is fine

PR #218, opened 2021-04-09
Upgrade the BOM generator to 0.0.10

aloubyansky:
Significantly reduces the time needed to generate the BOM.
It also configures explicitly the groupIds of the artifacts that should be checked for containing Quarkus extensions during the JSON descriptor generation.

### DIFF
--- a/descriptor/pom.xml
+++ b/descriptor/pom.xml
@@ -26,13 +26,13 @@
         <plugins>
             <plugin>
                 <groupId>io.quarkus</groupId>
-                <artifactId>quarkus-platform-descriptor-json-plugin</artifactId>
-                <version>${quarkus.version}</version>
+                <artifactId>quarkus-platform-bom-maven-plugin</artifactId>
+                <version>0.0.10</version>
                 <executions>
                     <execution>
                         <phase>process-resources</phase>
                         <goals>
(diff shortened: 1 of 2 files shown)


descriptor/pom.xml
```diff
@@ -26,20 +26,34 @@
...
+                        <groupId>com.blazebit</groupId>
+                        <groupId>io.quarkiverse.freemarker</groupId>
+                        <groupId>io.quarkiverse.googlecloudservices</groupId>
+                        <groupId>io.quarkiverse.minio</groupId>
+                        <groupId>io.quarkus</groupId>
+                        <groupId>org.apache.camel.quarkus</groupId>
+                        <groupId>io.debezium</groupId>
+                        <groupId>org.amqphub.quarkus</groupId>
+                        <groupId>com.datastax.oss.quarkus</groupId>
+                        <groupId>com.hazelcast</groupId>
```
aloubyansky, line 55: @gsmet we should be careful about this list when adding new members. We should probably introduce support for wildcards, e.g. `io.quarkiverse.*`, since anything coming from that group will include extensions.

gsmet: Yes support for wildcards would be nice.

gsmet (approved):

PR #220, opened 2021-04-15
Kogito 1.5.0 + OptaPlanner 8.5.0

evacchi:


paulrobinson:
@gsmet I added the `backport` label, as I think this is a candidate for 1.11.7.Final. Is this the right thing to do, given that this is the `quarkus-platform` repo?

It's requested in https://issues.redhat.com/browse/QUARKUS-867.

gsmet (approved):

gsmet:
Hmmm, you are aware that 1.11 is using Kogito 1.1 and OptaPlanner 8.1 so that's a big upgrade?

@evacchi could you aggregate all the PR upgrades since 1.11 for the 1.11 branch and create a PR if we really want to push this to 1.11? Thanks!

evacchi:
@gsmet I can do that, however QUARKUS-867 can be probably fixed by updating the quickstart to Kogito 1.1 :)

paulrobinson:
@evacchi that sounds safer, given that we are talking about backporting to a micro release. Do you want to check that it fixes the problem?

PR #257, opened 2021-06-30
Upgrade the BOM generator to 0.0.20

aloubyansky:
This adds gpg.skip=true to the test modules.


gsmet (approved):
Fast tracking it as I need it for the release!

PR #262, opened 2021-07-05
Updating Debezium to 1.6.0.Final

gunnarmorling:
@gsmet @geoand, quick Debezium version update here after our 1.6 Final release last week. There were no API changes whatsoever, I suppose that update in _pom.xml_ is all that's needed? Thanks!

### DIFF
--- a/generated-platform-project/quarkus-camel/bom/pom.xml
+++ b/generated-platform-project/quarkus-camel/bom/pom.xml
@@ -1927,6 +1927,11 @@
         <version>9.4</version>
       </dependency>
       <dependency>
+        <groupId>com.oracle.instantclient</groupId>
+        <artifactId>xstreams</artifactId>
+        <version>21.1.0.0</version>
+      </dependency>
+      <dependency>
         <groupId>com.orbitz.consul</groupId>
         <artifactId>consul-client</artifactId>
         <version>1.3.3</version>
@@ -2005,7 +2010,7 @@
       <dependency>
         <groupId>io.debezium</groupId>
         <artifactId>debezium-api</artifactId>
-        <version>1.6.0.CR1</version>
+        <version>1.6.0.Final</version>
       </dependency>
       <dependency>
         <groupId>io.debezium</groupId>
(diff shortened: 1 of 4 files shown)


geoand (approved):
Great, thanks!

aloubyansky:
We actually include the changes to the `generated-platform-project` as well to see the effects of a version upgrade on the whole platform. For now unfortunately that means building the platform locally with `mvn process-resources`. We'll be looking into automating this in the future.

gunnarmorling:
Ok, so what do I need to do exactly? Run `mvn process-resources` on this project, then what?

gsmet:
@gunnarmorling then you add the changes and you amend your commit with them.

gunnarmorling:
Ok, done and update pushed. One question for @Naros though: Is that newly added XStreams dependency from Maven Central?

aloubyansky:
> Ok, done and update pushed.

Thanks!

> One question for @Naros though: Is that newly added XStreams dependency from Maven Central? Or rather: why is this only showing up now for the Final, but not for the earlier CR release?

Actually, that's something I'll need to figure out. It doesn't look like your upgrade should be adding it.

gsmet:
@aloubyansky FWIW it's the second time it appears. Last time, it disappeared by itself with another PR.

aloubyansky:
Yes, I noticed, will have look tomorrow.

gunnarmorling:
Thanks for merging. So re that xstreams dependency; this one isn't available in any public repo, but instead the user would have to download it separately and install it into their own local Maven repo. Based on that, I think it'd be better to exclude this from the Quarkus platform BOM. Is there a way for you folks for blacklisting such particular dependency, @aloubyansky?

aloubyansky:
Sure, there is a way to exclude it https://github.com/quarkusio/quarkus-platform/blob/main/pom.xml#L396
The question to clarify is why it is added/removed. If it can't be resolved then it should be filtered out automatically. I'll look into this later today.

gunnarmorling:
So we have this as a provided dependency in Debezium, as this enables an optional feature of the Debezium connector for Oracle. But the user needs to download this JAR manually, it's not available in any public repo unfortunately. This isn't used a lot, and if so, people can still add the JAR to their Quarkus application, potentially also choosing another group id / artifact id. All i nall, I would suggest you automatically filter this out, your `bomgGenerator.excludedDependencies` dependencies looks like the right avenue for doing so. Thanks a lot!

aloubyansky:
Thanks for the info.

gsmet:
Backported to 2.0 branch for inclusion in 2.0.1.Final.

PR #270, opened 2021-07-15
Upgrade to Quarkus 2.1.0.CR1

gsmet:


gsmet:
@ppalaga @adutra FYI I had to disable the Cassandra and Camel tests for now as there is an incompatibility between Assertj versions. As soon as you upgrade your extensions to 2.1, we will be able to reenable them.

PR #277, opened 2021-07-27
Upgrade to Camel Quarkus 2.1.0

zbendhiba:


generated-platform-project/quarkus-debezium/bom/pom.xml
```diff
@@ -61,12 +61,12 @@
       <dependency>
         <groupId>io.debezium</groupId>
         <artifactId>debezium-quarkus-outbox-deployment</artifactId>
-        <version>1.6.0.Final</version>
+        <version>1.6.1.Final</version>
```
aloubyansky, line 64: I'll need to look into why Debezium got upgraded.

zbendhiba: From camel-quarkus bom

aloubyansky: The Camel Quarkus doesn't actually include these extensions, does it? https://repo1.maven.org/maven2/org/apache/camel/quarkus/camel-quarkus-bom/2.1.0/camel-quarkus-bom-2.1.0.pom
It is importing `io.debezium:debezium-bom:1.6.1` but it doesn't include these artifacts either https://repo1.maven.org/maven2/io/debezium/debezium-bom/1.6.1.Final/debezium-bom-1.6.1.Final.pom

It looks like the generator is aligning these extension versions to highest debezium version found in the platform.

Either way, one member (Camel) is not supposed to be overriding another member (Debezium) extension versions. The generator is supposed to be preventing that. I guess Debezium is a special case, since it's not providing any BOM as an input to the platform generator. Which is why I need to look into this.

aloubyansky: So, what I am saying here is that the Debezium as a member should be controlling its version, not Camel. What should happen in this specific case, the version of Debezium should remain 1.6.0.Final until this line changes https://github.com/quarkusio/quarkus-platform/blob/main/pom.xml#L60 Even if Camel Quarkus includes another (possibly a later) version.

zbendhiba: @aloubyansky Yes, I've hesitate to push that file. I didn't know which version of Debezium should be the promoted one, the one set by Debezium team, our by Camel Quarkus project.
What would have been the best way to handle that ?

aloubyansky: Don't worry about this. You did the right thing.

Every BOM in your PR was generated. The reason we are pushing all the changes under the `generated-platform-project` is to actually see the effects of the changes to the config in the root pom. Pushing the generated changes doesn't have any effect on the platform build, tests and release. Even if you didn't push the debezium BOM, it would still be updated on every platform build and in the release.

I've released a new version of the platform BOM generator that will prevent this kind of version override. The Debezium team has the right to control their version in the platform. I've already pinged them about this situation.

zbendhiba: Okay. Thanks @aloubyansky

PR #278, opened 2021-07-28
Upgrade to the platform BOM generator 0.0.24

aloubyansky:
This version makes sure members that do not provide BOMs as input to the platform generator don't get their extension versions overridden by members that do provide BOMs.


aloubyansky:
@zbendhiba just fyi, these are the effects of the new version of the generator and prevents Camel overriding the Debezium version.

aloubyansky:
It appears that Camel Quarkus is actually incompatible with the Debezium 1.6.0.Final.
````
Caused by: java.lang.NullPointerException
	at org.apache.camel.quarkus.component.debezium.common.it.AbstractDebeziumResource.receive(AbstractDebeziumResource.java:60)
````
I'll upgrade it to 1.6.1.Final.

PR #299, opened 2021-08-24
Removal of redhat version from quarkus versions

harshmadhani:
Removing the hard-coded redhat versions from the pom cc @aloubyansky


pom.xml
```diff
@@ -550,9 +550,9 @@
             <id>rhproduct</id>
             <properties>
                 <quarkus.upstream.version>${quarkus.version}</quarkus.upstream.version>
-                <quarkus-bom.version>${quarkus.upstream.version}-redhat-00001</quarkus-bom.version>
+                <quarkus-bom.version>${quarkus.upstream.version}</quarkus-bom.version>
                 <platform.groupId>com.redhat.quarkus.platform</platform.groupId>
-                <platform.version>${project.version}-redhat-00001</platform.version>
+                <platform.version>${project.version}</platform.version>
```
aloubyansky, line 555: @harshmadhani these two properties can be removed completely from this profile.
`platform.version` is already initialized to the `project.version` and the `quarkus-bom.version` should be set to a redhat version when you build it.

harshmadhani: But is not the quarkus-bom.version used as reference at some places in the pom?

aloubyansky: It is. You should not remove it completely from the pom but from the rhproduct profile.

aloubyansky: Only the property overrides.

PR #312, opened 2021-09-08
Updating JSON file to include jdbc-oracle support

harshmadhani:


harshmadhani:
cc @maxandersen

maxandersen:
@ia3andy can you confirm this matches what you expect ?

i  believe its fine to merge but just want to check you dont see any issue with committign it before code.quarkus.redhat.com is ready?

ia3andy:
@maxandersen looking good.. this will not be backported to 1.11 right?

geoand:
Is this good to backport to 2.2?

maxandersen:
Yes. Just 2.2 and onwards. So yes should be back ported to 2.2.3

PR #339, opened 2021-10-14
Kogito 1.12.0.Final + OptaPlanner 8.12.0.Final

radtriste:


aloubyansky:
@radtriste thanks for the update, is it targeting the 2.4.0.Final platform release?

evacchi:
if it's compatible with 2.4.0 then sure @aloubyansky but we won't be able to do a fresh release of 1.13 on time for that date probably

radtriste:
> @radtriste was the Jenkins pipeline clean?

the release pipeline was all clean with Quarkus 2.3
Maybe we should just change the target branch and see if better. I can update the PR for that.

Did not try to run the platform test locally though. This is quite automated.

evacchi:
we released with 2.3. We were trying to target whatever Quarkus.next is (2.4.x in this case)

aloubyansky (approved):

PR #360, opened 2021-11-01
Disable Camel Spring RabbitMQ tests

jamesnetherton:
As I mentioned [here](https://github.com/quarkusio/quarkus/issues/8593#issuecomment-956009053), I think in the long term it'll be better to start testing with CQ SNAPSHOTS so that we don't have to constantly enable / disable tests.


aloubyansky:
Thanks @jamesnetherton It makes sense. Although we will still want to test the current platform with the Quarkus main branch to check whether changes in Quarkus break the current platform.

PR #369, opened 2021-11-12
Platform generator 0.0.36

aloubyansky:
This upgrade includes a few changes:
* it drops filtering of what appears to be constraints managed by other members from a member BOM;
* it detects all the platform BOM imports in the original member BOMs and subtracts all of them instead of only detected `io.quarkus:quarkus-bom` imports;
* introduces an option to disabled BOM report generation and an option to create a ZIP containing the reports.

This PR also includes a commit adding `sync` Maven profile that is a convenience to quickly sync the `generated-platform-project` with the platform config in the root `pom.xml`. So, instead of `mvn clean process-resources` previously, `mvn -Dsync` can be used before pushing changes and opening a PR.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -92,4 +92,4 @@ jobs:
         if: ${{ steps.get-changes.outputs.changes > 0 }}
         with:
           script: |
-            throw new Error('Generated project needs an update. Please run ./mvnw clean process-resources and include the result in your pull request.')
+            throw new Error('Generated project needs an update. Please run ./mvnw -Dsync and include the result in your pull request.')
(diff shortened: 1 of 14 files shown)


ppalaga (approved):

ppalaga:
Thank @aloubyansky !

gsmet (approved):

gsmet:
Maybe you could adjust the hint here: https://github.com/quarkusio/quarkus-platform/blob/main/.github/workflows/build.yml#L95 ?

aloubyansky:
Good idea, done, thanks @gsmet

PR #371, opened 2021-11-15
Upgrade to the platform generator 0.0.37

aloubyansky:
This release disables the alignment of member own constraints if no other member includes constraints that create a conflict. (This also reduces the platform generation time). The previous behavior can be enabled by adding `<alignOwnConstraints>true</alignOwnConstraints>` element to a member's config.

For example, Camel Quarkus imports the Azure SDK BOM and no other member includes these constraints. This release will not attempt to align them by default.

FYI @ppalaga @jamesnetherton @vkasala

### DIFF
--- a/generated-platform-project/quarkus-camel/bom/pom.xml
+++ b/generated-platform-project/quarkus-camel/bom/pom.xml
@@ -121,12 +121,12 @@
       <dependency>
         <groupId>com.azure</groupId>
         <artifactId>azure-communication-common</artifactId>
-        <version>1.0.2</version>
+        <version>1.0.1</version>
       </dependency>
       <dependency>
         <groupId>com.azure</groupId>
         <artifactId>azure-communication-identity</artifactId>
-        <version>1.1.1</version>
+        <version>1.1.0</version>
       </dependency>
       <dependency>
         <groupId>com.azure</groupId>
@@ -236,17 +236,17 @@
       <dependency>
         <groupId>com.azure</groupId>
         <artifactId>azure-security-keyvault-keys</artifactId>
-        <version>4.3.0</version>
+        <version>4.2.8</version>
       </dependency>
       <dependency>
         <groupId>com.azure</groupId>
         <artifactId>azure-security-keyvault-secrets</artifactId>
-        <version>4.3.0</version>
+        <version>4.2.8</version>
       </dependency>
       <dependency>
         <groupId>com.azure</groupId>
         <artifactId>azure-storage-blob-batch</artifactId>
-        <version>12.11.1</version>
+        <version>12.9.1</version>
       </dependency>
       <dependency>
         <groupId>com.azure</groupId>
@@ -271,7 +271,7 @@
       <dependency>
         <groupId>com.azure</groupId>
         <artifactId>azure-storage-file-share</artifactId>
-        <version>12.11.1</version>
+        <version>12.9.1</version>
       </dependency>
       <dependency>
         <groupId>com.azure</groupId>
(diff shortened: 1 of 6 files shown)


generated-platform-project/quarkus-camel/bom/pom.xml
```diff
@@ -647,17 +647,17 @@
       <dependency>
         <groupId>io.dropwizard.metrics</groupId>
         <artifactId>metrics-healthchecks</artifactId>
-        <version>4.1.19</version>
+        <version>4.0.1</version>
```
aloubyansky, line 650: This change (and other dropwizard, org.graalvm.js and apache.curator) indicate that your BOM includes other artifacts that originate from different releases of the same project. I.e. they were previously aligned by the platform generator but now they are not.

ppalaga: This is a bit strange.

For dropwizard and Camel Quarkus 2.4.0, I'd expect the version 4.1.25, defined in `org.apache.camel:camel-dependencies:3.12.0`

Similarly `org.graalvm.js` should be at `21.2.0` see https://github.com/apache/camel-quarkus/blob/2.4.0/poms/bom/pom.xml#L6084 and https://github.com/apache/camel-quarkus/blob/2.4.0/pom.xml#L83

aloubyansky: You have two artifacts coming from `org.graalvm.js`: `js` one is `21.2.0` but the `js-scriptengine` is `20.0.0`.
As to the dropwizard, you should check but that's what they actually are. An easy way to verify it is to download https://repo1.maven.org/maven2/org/apache/camel/quarkus/camel-quarkus-bom/2.4.0/camel-quarkus-bom-2.4.0.pom as `pom.xml` and run `mvn help:effective-pom > effective-pom.xml`. You'll see the flattened version of your BOM with all the actual constraints.

ppalaga: Going to check that now

ppalaga: You are right, the BOM generator works as designed, we need to be more careful what we manage in our BOM. I filed https://github.com/apache/camel-quarkus/issues/3302 for that. 
These insights are super useful for us - thanks a lot for them!

ppalaga (approved):

ppalaga:
BTW, in the prod platform 2.2.x, I'll be for having `<alignOwnConstraints>true</alignOwnConstraints>` and keeping the enforcedDependencies

```
                            <enforcedDependencies>
                                <!-- Workaround for https://issues.redhat.com/browse/QUARKUS-1379 -->
                                <dependency>com.azure:azure-storage-blob-batch:12.9.1</dependency>
                                <dependency>com.azure:azure-storage-file-share:12.9.1</dependency>
                                <dependency>com.azure:azure-storage-queue:12.9.1</dependency>
                            </enforcedDependencies>
```

because we do not want to rebuild Camel Quarkus.

ppalaga:
Would you please ping me when you have backported this to the prod branch?

gsmet:
I wonder if we should merge this until things are fixed in Camel Quarkus?

ppalaga:
> I wonder if we should merge this until things are fixed in Camel Quarkus?

https://github.com/apache/camel-quarkus/pull/3303 contains the minimal set of fixes and I hope it will be merged tomorrow. So a fixed Camel Quarkus BOM should be ready for Platform 2.5.0.Final

gsmet:
OK, no objections from me in this case.

ppalaga:
https://github.com/apache/camel-quarkus/pull/3303 is merged now.

PR #373, opened 2021-11-17
Update to Quarkus core 2.5.0.Final

gemmellr:
Simple version change and build then commit. Done to isolate out the Core changes from those coming later in a PR updating a component. Figured it made sense to just raise the PR with them while working on the rest.


gsmet (approved):
Nice, thanks!

PR #382, opened 2021-11-24
Kogito 1.14.0.Final + OptaPlanner 8.14.0.Final

evacchi:


gsmet:
Please run the sync with the `./mvnw` wrapper (``./mvnw -Dsync`). They changed the order of elements in the exclusions and it generates a lot of noise...

We will upgrade to 3.8.4 in a few days but I'll make it a separate commit.

evacchi:
ah so that's why

PR #415, opened 2021-12-23
Upgrade Maven Wrapper to 3.1.0

jorsol:
I have run `./mvnw -Dsync` but it overwrites the version from 
```
generated-platform-project/quarkus-universe/descriptor/src/main/resources/overrides.json
generated-platform-project/quarkus/descriptor/src/main/resources/overrides.json
```
So I'm not sure if I should skip those files and only update the wrapper for the project.


gsmet:
Can you push the result of `./mvnw -Dsync` so that we can have a look?

jorsol:
The result is to keep `"maven-wrapper-version" : "0.7.7",` my guess is that the platform generate those files from `<quarkus.version>2.6.0.Final</quarkus.version>` and since the "new" wrapper is found in `main` (2.7.0) then this will sync to the version used in 2.6.0.Final.

So, this should only update the wrapper used in the repo, I will run -Dsync again and just push the wrapper update without the overrides.json.

aloubyansky:
Please, always include all the changes to the `generated-platform-project`. This project is re-generated on every build and the whole purpose of it in github is to see how it is changing.

jorsol:
I might be missing something, but there are no changes to the `generated-platform-project` (-Dsync).

aloubyansky:
That's ok, I thought there were and you tried to not include them. Thanks.

gsmet (approved):

gsmet:
Thanks!

PR #457, opened 2022-01-28
Google Cloud services extension pack 1.0.0

loicmathieu:
When I launchd `mvn -Dsyn` it updates more than 200 pom with artifactid and groupId inverted !

I didn' commit all of them so I manually include only the changes made to version and new libraires on the 4 files that are usually udated when I update the platform for the Google Cloud services extensions.

As I may miss some changes in one of those 200 pom, maybe it would be better if someone can validate my PR locally (or just sync the platform after merging it).

Sorry for that :(

### DIFF
--- a/generated-platform-project/quarkus-camel/bom/pom.xml
+++ b/generated-platform-project/quarkus-camel/bom/pom.xml
@@ -422,12 +422,12 @@
       <dependency>
         <groupId>com.google.api.grpc</groupId>
         <artifactId>proto-google-common-protos</artifactId>
-        <version>2.6.0</version>
+        <version>2.7.1</version>
       </dependency>
       <dependency>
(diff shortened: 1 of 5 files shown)


aloubyansky:
@loicmathieu it's probably because you used `mvn` instead of the wrapper from the project.

loicmathieu:
Well, I never use the wrapper ...

Let's wait for Monday so I can launch with the wrapper and be sure to catch all modifications.

loicmathieu:
And looking at the CI status there is missing modifications.

gsmet:
For the record: they changed how things were generated between two micro versions of Maven. That's why we use the wrapper now to make sure things stay consistent.

gsmet (approved):

gsmet:
Thanks!

PR #478, opened 2022-02-21
Upgrade Quarkus Vault to 1.0.2

gsmet:
@vsevel FYI, Vault is included in the Platform so we need to upgrade it here when we push a new version (especially if it's to fix something broken).

Once, the upgrade is done, you need to run `./mvnw -Dsync`, wait and commit everything.

### DIFF
--- a/generated-platform-project/quarkus-qpid-jms/integration-tests/quarkus-qpid-jms-integration-tests/pom.xml
+++ b/generated-platform-project/quarkus-qpid-jms/integration-tests/quarkus-qpid-jms-integration-tests/pom.xml
@@ -27,7 +27,7 @@
     <dependency>
       <groupId>io.quarkiverse.artemis</groupId>
       <artifactId>quarkus-test-artemis</artifactId>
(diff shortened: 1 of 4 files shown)


generated-platform-project/quarkus-qpid-jms/integration-tests/quarkus-qpid-jms-integration-tests/pom.xml
```diff
@@ -27,7 +27,7 @@
     <dependency>
       <groupId>io.quarkiverse.artemis</groupId>
       <artifactId>quarkus-test-artemis</artifactId>
-      <version>1.0.1</version>
+      <version>${quarkus-config-consul.version}</version>
```
gsmet, line 30: This is just an artifact. Apparently, it uses a property if only one property points to this particular version.

aloubyansky: Correct

PR #507, opened 2022-03-30
Upgrade to Quarkus 2.8.0.Final

gsmet:


gsmet:
There is a sync issue with the Maven repo. This PR is not right as it couldn't get all the updated files from Maven Central.

I'll check in a couple of hours.

gemmellr:
@gsmet I was able to complete a `./mvnw clean install -DskipTests` using 2.8.0.Final just now.

gemmellr:
Though as I mentioned on the sonatype JIRA, the https://repo1.maven.org/maven2/io/quarkus/quarkus-webjars-locator/ module is still inconsistent for me, so the platform update I can generate locally is actually missing an entry for that from the boms (the deployment bit is ok), i.e the generation actually removed it.

Though interestingly, I also pushed my changes to GitHub earlier and the build on my fork _was_ able to generate the correct changes in CI, and thus failed the check job as it generated a different output than I had pushed, with the removed bit then present.

gemmellr:
It occurs to me that the CI job is using the Google mirror of Central earlier, while locally I wasnt...depending on how that mirror syncing is done, it seems possible the mirror might have been correct from the prior prodding sonatype did, even while Central's CDN cache was still stale/wrong. Now that they prodded the CDN cache for quarkus-webjars-locator to get it working, I was able to get a local build that matched what CI generated, and passes the comparison check job.

I have raised PR #512 with that change and my Quarkus Qpid JMS update. I can rebase it if you want to put the 2.8.0.Final change in via this PR.

PR #521, opened 2022-04-07
Configure Cassandra's own groupIds

aloubyansky:
@ppalaga I wanted to discuss this with you. Here I am configuring groupIds that belong to Cassandra and that other members aren't allowed to override versions for.
The way it works currently, it takes the complete dependency element from the Cassandra BOM and replicates everywhere, including the Camel BOM. Which is why you see the exclusions being removed in from the Camel BOM. In fact, I can preserve inclusions in the Camel BOM but it seems like none of the approaches will be always right. I think the current behavior has more chances to be right in the majority of cases.

### DIFF
--- a/generated-platform-project/quarkus-camel/bom/pom.xml
+++ b/generated-platform-project/quarkus-camel/bom/pom.xml
@@ -284,22 +284,12 @@
             <groupId>org.hdrhistogram</groupId>
             <artifactId>HdrHistogram</artifactId>
           </exclusion>
-          <exclusion>
-            <groupId>com.google.code.findbugs</groupId>
-            <artifactId>jsr305</artifactId>
-          </exclusion>
         </exclusions>
       </dependency>
       <dependency>
         <groupId>com.datastax.oss</groupId>
         <artifactId>java-driver-query-builder</artifactId>
         <version>4.14.0</version>
-        <exclusions>
-          <exclusion>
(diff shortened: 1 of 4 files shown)


ppalaga:
> I am configuring groupIds that belong to Cassandra and that other members aren't allowed to override versions for.

+1 on the idea :-)

> The way it works currently, it takes the complete dependency element from the Cassandra BOM and replicates everywhere, including the Camel BOM. Which is why you see the exclusions being removed in from the Camel BOM. 

findbugs specifically is not a big issue for us (as long as it is managed at some reasonable version, which it is since some time). So this PR is good to go.

There might be other exclusions (such as old specs) that may matter more to us in other future situations similar to this one. For those, it would be nice if Quarkus had some mechanism for sharing bans and dependency replacements. Something like 
* `org.apache.geronimo:spec-foo:*` is banned in extensions, tests, and BOMs (MUST be excluded)
* If such banned artifact is present, a participant's build MUST fail
* If such banned artifact is transitively present in participant's BOM, the given participant version cannot be allowed in the platform or must be sanitized during Platform BOM generation
* Each such ban MAY optionally contain a replacement artifact (such as org.jakarta:jakarta-foo), that (1) MUST be managed and (2) MUST be present in every extension where the original artifact is excluded.

It is vital that all platform participants comply with the same set of bans and replacements.

aloubyansky:
Thanks @ppalaga

PR #530, opened 2022-04-20
Disable Camel JPA tests

jamesnetherton:
Relates to this failure on the nightly build:

https://github.com/quarkusio/quarkus/issues/8593#issuecomment-1098842106


aloubyansky:
@jamesnetherton it failed on other tests now, could you have a look?

jamesnetherton:
> @jamesnetherton it failed on other tests now, could you have a look?

I've never seen those tests fail before. Perhaps a temporary glitch. Can we rerun the workflow?

If it becomes a frequent source of failures, I'll disable those tests and investigate it.

aloubyansky:
Sure, re-running.

PR #555, opened 2022-06-01
Kogito 1.22.1.Final + OptaPlanner 8.22.1.Final

mbiarnes:

Thanks!


aloubyansky:
Re-running

aloubyansky (approved):
Thanks!

aloubyansky:
Do you want this in 2.9.3.Final? Or only 2.10.x?

radtriste:
if we can have it in 2.9.3 that would be great !

aloubyansky:
Ah, sorry, I didn't realize that. Then it'll target 2.10.x, I suppose. BTW, Guillaume is on PTO, in case you notice him being silent.

radtriste:
ok then for 2.10

PR #557, opened 2022-06-07
Disable Camel Google Storage tests

jamesnetherton:
Relates to https://github.com/quarkusio/quarkus/issues/8593#issuecomment-1148271720.

This will probably need backporting to any other active branches unfortunately.


jamesnetherton:
Weird - the issue seems to have resolved itself. I'll close this for now.

There's a change coming for this test in CQ 2.10.0 that will improve things if the same issue pops up in future.

PR #560, opened 2022-06-20
Make software.amazon.awssdk groupId owned by the AmazonServices

aloubyansky:
Make software.amazon.awssdk groupId owned by the AmazonServices to avoid a possibiliy for other members to override versions of these dependencies

FYI @ppalaga

### DIFF
--- a/generated-platform-project/quarkus-amazon-services/bom/pom.xml
+++ b/generated-platform-project/quarkus-amazon-services/bom/pom.xml
@@ -222,12 +222,6 @@
         <groupId>software.amazon.awssdk</groupId>
         <artifactId>apache-client</artifactId>
         <version>2.17.166</version>
-        <exclusions>
-          <exclusion>
-            <groupId>commons-logging</groupId>
-            <artifactId>commons-logging</artifactId>
-          </exclusion>
-        </exclusions>
       </dependency>
       <dependency>
         <groupId>software.amazon.awssdk</groupId>
--- a/generated-platform-project/quarkus-camel/bom/pom.xml
+++ b/generated-platform-project/quarkus-camel/bom/pom.xml
@@ -526,12 +526,22 @@
             <groupId>org.hdrhistogram</groupId>
             <artifactId>HdrHistogram</artifactId>
           </exclusion>
+          <exclusion>
+            <groupId>com.google.code.findbugs</groupId>
+            <artifactId>jsr305</artifactId>
+          </exclusion>
         </exclusions>
       </dependency>
       <dependency>
         <groupId>com.datastax.oss</groupId>
         <artifactId>java-driver-query-builder</artifactId>
         <version>4.14.0</version>
+        <exclusions>
+          <exclusion>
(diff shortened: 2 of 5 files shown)


generated-platform-project/quarkus-amazon-services/bom/pom.xml
```diff
@@ -222,12 +222,6 @@
         <groupId>software.amazon.awssdk</groupId>
         <artifactId>apache-client</artifactId>
         <version>2.17.166</version>
-        <exclusions>
-          <exclusion>
-            <groupId>commons-logging</groupId>
-            <artifactId>commons-logging</artifactId>
-          </exclusion>
-        </exclusions>
```
ppalaga, line 230: Could you please add back the exclusions in the BOM generator config?

aloubyansky: The way I decided to do it is to add an option that would allow each member to keep their own exclusions for common dependencies. While aligning the common dependency version is critical, I am not sure whether it's a good idea to allow one member to control exclusions for others, especially if that member didn't do a proper research on that. So I added this, probably, controversial option `keepThirdpartyExclusions` element to member config (which is `true` by default) that allows members to keep their exclusions configured in their original BOMs.
This could make the order of member BOM imports in Quarkus apps significant, although that'll probably be an edge case.
So in this particular change the exclusion from the `quarkus-amazon-services` BOM is removed but for this same constraint in the `quarkus-camel-bom` it is still present.
WDYT @ppalaga and @gsmet?

maxandersen: Not following why we wouldn't want alignment around exclusions be consistent across the platform ?

Users can add explicit exclusions/inclusions if they somehow need it for their usecase, right ?

aloubyansky: I am not sure how to do that automatically so everyone is happy about that. If you have a suggestion, please share. We can configure agreed upon exclusions in the platform config today, it'd be a static config though that would need to be adjusted on changes to the member BOMs, which will be difficult to track.

maxandersen: Couldn't we have the tool apply all exclusions and fail platform build if items added that isn't in a agreed upon list ?

aloubyansky: It wouldn't be much different from simply adding a common dependency constraint with exclusions for all the members, right? This can be done today in the platform config.

aloubyansky: Sorry, I am going to merge it, since I need to upgrade the platform generator for other reasons. Please create an issue if you think we could have a more appropriate default.

ppalaga: I have filed a follow up https://github.com/quarkusio/quarkus-platform-bom-generator/issues/114

PR #565, opened 2022-06-22
Disable temporarely kogito 1.23 failing tests

radtriste:

Thanks!


radtriste:
@aloubyansky not sure why the check failed ...
I run `./mvnw -Dsync` on my local but there is no change ...

PR #571, opened 2022-07-04
kogito 1.24.0.Final

mbiarnes:

Thanks!

Contains staging repos that should be removed once the Kogito and OptaPlanner artifacts are in Maven Central


radtriste:
@aloubyansky @gsmet optaplanner is using now Drools 8.23.0.Beta and kogito is using Drools 8.24.0.Beta
Why is the sync command not putting the 8.24.0.Beta ? 
Is there a way to force it ?

aloubyansky:
Because only the optaplanner BOM includes the `org.drools` constraints. There is no conflict from the dependency version aligning mechanism.
Either you should include the `org.drools` constraints in the kogito BOM or configure them explicitly in the platform config.
But you can't use different versions for Kogito and OptaPlanner.

radtriste:
I think we should include the `org.drools` constraints in the kogito BOM in the next version ... cc @evacchi
For now, is there a simple way to override the drools used version in the platform config ? or do we have to specify all dependencies as configuration ?

evacchi:
I am honestly not sure, how does OptaPlanner does it? @rsynek ? /cc @mariofusco

aloubyansky:
Yes, there is an easy way to upgrade the Drools version for OptaPlanner in the platform, by adding e.g.:
````
diff --git a/pom.xml b/pom.xml
index bd8c84eb..1fbc35fc 100644
--- a/pom.xml
+++ b/pom.xml
@@ -282,6 +282,9 @@
                                     <lastDetectedBomUpdate>io.quarkus.platform:quarkus-kogito-bom:2.10.0.CR1</lastDetectedBomUpdate>
                                     <next>${platform.groupId}:quarkus-kogito-bom:${platform.version}</next>
                                 </release>
+                                <dependencyManagement>
+                                    <dependency>org.drools:drools-beliefs:8.24.0.Beta</dependency>
+                                </dependencyManagement>
````

PR #574, opened 2022-07-05
optaplanner 8.24.0.Final

mbiarnes:

Thanks!


gsmet:
Is there a reason why you don't upgrade both OptaPlanner and Kogito in the same PR as usual? It looks better to me.

mbiarnes:
Kogito and Optaplanner have now separate releases - they have different release cycles now.

PR #587, opened 2022-07-21
Add camel-quarkus-test-list.version property

aloubyansky:
This allows to control the test list artifact version using a separate property.


aloubyansky:
FYI @ppalaga @jamesnetherton I would like to be able to control the test catalog version separately from the Camel Quarkus version.

PR #606, opened 2022-08-17
Upgrade to Quarkus 2.12.0.CR1

gsmet:


pom.xml
```diff
@@ -277,6 +277,10 @@
                                             <project.groupId>${project.groupId}</project.groupId>
                                         </systemProperties>
                                     </defaultTestConfig>
+                                    <!--
+                                    Build step 'org.kie.kogito.infinispan.quarkus.deployment.KogitoAddOnPersistenceInfinispanProcessor#verifyCapabilities' does not produce any build item and thus will never get executed
+                                    -->
```
gsmet, line 282: @evacchi @cristianonicolai @mbiarnes this commit will have to be reverted when we upgrade Kogito.

evacchi: /cc @radtriste

radtriste: oki
cc @mbiarnes

radtriste: This would be for Kogito 1.27 then

PR #614, opened 2022-08-30
Vault extension 2.0.0

vsevel:
tested with 2.12.0
thanks a lot @kdubb 
cc @aloubyansky

### DIFF
--- a/generated-platform-project/quarkus-universe/bom/pom.xml
+++ b/generated-platform-project/quarkus-universe/bom/pom.xml
@@ -6267,17 +6267,17 @@
       <dependency>
         <groupId>io.quarkiverse.vault</groupId>
         <artifactId>quarkus-vault-deployment</artifactId>
-        <version>2.0.0-RC1</version>
+        <version>2.0.0</version>
       </dependency>
       <dependency>
         <groupId>io.quarkiverse.vault</groupId>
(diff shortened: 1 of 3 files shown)


aloubyansky (approved):

gsmet:
Ah, cool, just in time :).

kdubb:
I'm gonna say that is an unrelated failure.

aloubyansky:
FYI @loicmathieu here is another PR (with unrelated changes) where it fails.

loicmathieu:
@aloubyansky is it the only test that fails from time to time or does other platform participant have dangling tests ?
If so, maybe we propose a tag to add to test that should not be run by the platform. I agree it's not a good solution but it allows to blacklist a test and offer time to fix it (at the risk of never fixing it)

aloubyansky:
@loicmathieu you can surely skip a test, for example https://github.com/quarkusio/quarkus-platform/blob/main/pom.xml#L419-L422 I just wanted to let you know there is an issue.

loicmathieu:
@aloubyansky if I didn't find the issue, I'll disable the test and open an issue on the Quarkiverse repo to avoid forgetting about it.

aloubyansky:
Perfect @loicmathieu, thanks!

PR #623, opened 2022-09-20
Enable not owned Quarkiverse extensions in member BOMs

aloubyansky:
This PR enables adding Quarkiverse extensions not owned by members in their member BOMs. Enabling this option revealed a conflict between CEQ and Kogito on the `io.quarkiverse.jackson-jq`. While CEQ depends on version `1.1.0`, Kogito depends on `1.0.1`. I had to downgrade it to `1.0.1` to be able to build the platform w/o errors.

FYI info @ppalaga and @radtriste 

Just FYI, earlier (relatively long ago) Quarkiverse not owned by members were by default excluded from platform BOMs to let Quarkiverse extensions not be overshadowed by the platform extensions in the extension catalog and project dependency management. Now that the platform generator and the platform itself has advanced, we could reconsider adding Quarkiverse extensions to member BOMs.


ppalaga:
Thanks for the fix, @aloubyansky ! Is there Kogito issue for upgrading to jackson-jq 1.1?

aloubyansky:
I haven't created one, no. I wanted to bring this up with the Kogito team.
I think going forward, I will switch the default for `enableNonMemberQuarkiverseExtensions` to `true` to not require an explicit config for it, just fyi.

ppalaga:
@radtriste any chance to take care for the upgrade?

radtriste:
@ppalaga sorry for the delay.
@cristianonicolai would you have time to look into it ?

aloubyansky:
I created https://issues.redhat.com/browse/KOGITO-8007

cristianonicolai:
Fix in progress https://github.com/kiegroup/kogito-runtimes/pull/2507

PR #652, opened 2022-10-13
Update resteasy-mutiny and reactive routes support scope

cescoffier:
- RESTEasy Mutiny (server and client) are deprecated -> move to RESTEasy reactive
- Reactive Routes are deprecated -> move to RESTEasy reactive


src/main/resources/core/extensions-support-overrides.json
```diff
@@ -492,7 +492,7 @@
       "artifact-id": "quarkus-reactive-routes",
       "metadata": {
         "redhat-support": [
-          "supported"
```
aloubyansky, line 495: @cescoffier are you sure this should be under "redhat-support"? I thought it's either supported or not (possibly as a tech-preview). `deprecated` is a tag that could be applied in the community catalog and it will still show up on code.quarkus.redhat.com next to the redhat-support tag.
@maxandersen @tqvarnst wdyt about it?

aloubyansky: For example https://github.com/quarkusio/quarkus/blob/main/extensions/reactive-routes/runtime/src/main/resources/META-INF/quarkus-extension.yaml
could change status to `deprecated` as it's done here
 https://github.com/quarkusio/quarkus/blob/main/extensions/resteasy-classic/rest-client-mutiny/runtime/src/main/resources/META-INF/quarkus-extension.yaml#L14

cescoffier: It is supported, and now we deprecate it.

cescoffier: (But it is still supported)

cescoffier: I see; I've removed "supported".

src/main/resources/core/extensions-support-overrides.json
```diff
@@ -537,7 +537,7 @@
       "artifact-id": "quarkus-rest-client-mutiny",
```
rsvoboda, line 537: this entry should be dropped, no deprecation needed from TP

src/main/resources/core/extensions-support-overrides.json
```diff
@@ -609,7 +609,7 @@
       "artifact-id": "quarkus-resteasy-mutiny",
```
rsvoboda, line 609: this entry should be dropped, no deprecation needed from TP

rsvoboda (changes requested):

rsvoboda:
@cescoffier could you refocus this PR just to `quarkus-rest-client-mutiny` and `quarkus-resteasy-mutiny`?
Just remove the entries for them, we are dropping TP support for them, so no product metadata are required.

reactive routes need decision first and it will probably take some time as there is no response in email thread yet.

rsvoboda:
looking at pending "DROP XYZ" JIRAs, I think I will create dedicated PR for that, so no action on this PR / from your side @cescoffier is needed atm

rsvoboda:
mutiny metadata drop is in https://github.com/quarkusio/quarkus-platform/pull/666 + https://github.com/quarkusio/quarkus-platform/pull/667

src/main/resources/core/extensions-support-overrides.json
maxandersen: > `deprecated` is a tag that could be applied in the community catalog and it will still show up on code.quarkus.redhat.com next to the redhat-support tag.

can you specific what you refer to here - "deprecated" in "status" key is separate from "deprecated" in "redhat-supported" - I would definitely recommend we don't mix those up if can be avoided.

maxandersen: note, this is most likely caused by code.quarkus limitation that "flattens" the model into one set of tags which is unfortunate :/ all other devtools are able to differentiate.

aloubyansky (approved):

PR #668, opened 2022-10-25
Enable Camel Kafka Oauth test

zbendhiba:

Thanks!

### DIFF
--- a/src/main/resources/xslt/camel/test-pom.xsl
+++ b/src/main/resources/xslt/camel/test-pom.xsl
@@ -112,11 +112,24 @@
         </xsl:copy>
     </xsl:template>
 
-  <!-- Forward solr.trust-store to the tests via surefire plugin -->
+  <!-- Add system property variables needed in Solr & Kafka Oauth Native Integration test -->
     <xsl:template match="//*[local-name() = 'systemPropertyVariables']">
         <xsl:copy>
             <xsl:if test="/pom:project/pom:artifactId/text() = 'camel-quarkus-integration-test-solr'">
                 <javax.net.ssl.trustStore>${solr.trust-store}</javax.net.ssl.trustStore>
+            </xsl:if>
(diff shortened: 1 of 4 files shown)


zbendhiba:
I just realized I need to add the System properties in https://github.com/quarkusio/quarkus-platform/blob/main/src/main/resources/xslt/camel/test-pom.xsl

Let me update this

zbendhiba:
It seems I triggered too many CI builds, sorry about that

src/main/resources/xslt/camel/test-pom.xsl
```diff
@@ -112,12 +112,25 @@
         </xsl:copy>
     </xsl:template>
 
-  <!-- Forward solr.trust-store to the tests via surefire plugin -->
+  <!-- Add system property variables needed in Solr & Kafka Oauth Native Integration test -->
```
aloubyansky, line 115: Is there a reason to do it this way instead of configuring  `systemProperties` under the corresponding `test` element?
For example:
````
<test>
    <artifact>org.apache.camel.quarkus:camel-quarkus-integration-test-kafka-oauth:${camel-quarkus-test-list.version}</artifact>
    <systemProperties>
        <javax.net.ssl.trustStore>${solr.trust-store}</javax.net.ssl.trustStore>
    </systemProperties>
</test>
````

zbendhiba: No this was my first attempt. Should I ? It should be easier. Yet, I don' know why it's still failing

zbendhiba: I did that in one of the previous builds but it failed too.
I guess I need to move out the file from target to have it running here.
I'll close this PR, and try to do changes for next release. And see how it goes

aloubyansky: Let me know if I can help with the test config.

PR #672, opened 2022-10-26
Add separate version property for Cassandra test artifacts

jamesnetherton:
/CC @aloubyansky as per our discussion.

### DIFF
--- a/generated-platform-project/quarkus-cassandra/integration-tests/cassandra-quarkus-integration-tests-dse/pom.xml
+++ b/generated-platform-project/quarkus-cassandra/integration-tests/cassandra-quarkus-integration-tests-dse/pom.xml
@@ -10,9 +10,6 @@
   </parent>
   <artifactId>cassandra-quarkus-integration-tests-dse</artifactId>
   <name>Quarkus Platform - Cassandra - Integration Tests - cassandra-quarkus-integration-tests-dse</name>
-  <properties>
-    <maven.test.skip>true</maven.test.skip>
-  </properties>
   <dependencies>
     <dependency>
       <groupId>com.datastax.oss.quarkus</groupId>
@@ -163,7 +160,6 @@
                   <goal>build</goal>
                 </goals>
                 <configuration>
-                  <skip>true</skip>
                   <appArtifact>com.datastax.oss.quarkus:cassandra-quarkus-integration-tests-dse:${quarkus-cassandra-client.version}</appArtifact>
                 </configuration>
               </execution>
--- a/generated-platform-project/quarkus-cassandra/integration-tests/cassandra-quarkus-integration-tests-main/pom.xml
+++ b/generated-platform-project/quarkus-cassandra/integration-tests/cassandra-quarkus-integration-tests-main/pom.xml
@@ -10,9 +10,6 @@
   </parent>
   <artifactId>cassandra-quarkus-integration-tests-main</artifactId>
   <name>Quarkus Platform - Cassandra - Integration Tests - cassandra-quarkus-integration-tests-main</name>
-  <properties>
-    <maven.test.skip>true</maven.test.skip>
-  </properties>
   <dependencies>
     <dependency>
       <groupId>com.datastax.oss.quarkus</groupId>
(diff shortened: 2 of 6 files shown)


jamesnetherton:
Just realized I missed tweaking the version in `<appArtifact>` for the `quarkus-maven-plugin` config. I'll work on fixing that.

jamesnetherton:
> Just realized I missed tweaking the version in `<appArtifact>` for the `quarkus-maven-plugin` config. I'll work on fixing that.

Should be fixed now.

src/main/resources/xslt/cassandra/test-pom.xsl
```diff
@@ -0,0 +1,26 @@
...
+
+    <!-- Force quarkus-cassandra-client-test.version property for any artifactId matching cassandra-quarkus-integration-tests -->
+    <xsl:template match="/pom:project/pom:dependencies/pom:dependency/pom:version">
+        <xsl:copy>
+            <xsl:if test="starts-with(../pom:artifactId/text(), 'cassandra-quarkus-integration-tests')"><xsl:apply-templates select="@*" />${quarkus-cassandra-client-test.version}</xsl:if>
+        </xsl:copy>
+    </xsl:template>
+    <xsl:template match="/pom:project/pom:profiles/pom:profile/pom:build/pom:plugins/pom:plugin[2]/pom:executions/pom:execution/pom:configuration/pom:appArtifact">
+        <xsl:copy><xsl:apply-templates select="@*" />com.datastax.oss.quarkus:<xsl:value-of select="/pom:project/pom:artifactId/text()"/>:${quarkus-cassandra-client-test.version}</xsl:copy>
+    </xsl:template>
```
aloubyansky, line 24: @jamesnetherton is this really necessary? Looks like it shouldn't be.

jamesnetherton: Not 100% sure TBH. I just saw that the `<appArtifact>` thing referenced the plain `quarkus-cassandra-client.version` property and thought it should probably point at `quarkus-cassandra-client-test.version` instead.

aloubyansky: Thanks for checking, I'll have a look.

aloubyansky: @jamesnetherton that's fine, you don't need to use XSLT here. It's just it's trying to guess the property here and is picking up the Cassandra BOM one for it. Try changing the test version to something that doesn't match the BOM version and you'll see it using the proper one for the tests.

jamesnetherton: Thanks for clarifying @aloubyansky. I have removed that part of the XSLT.

src/main/resources/xslt/cassandra/test-pom.xsl
```diff
@@ -0,0 +1,22 @@
...
+        <xsl:copy>
+            <xsl:apply-templates select="@* | node()" />
+        </xsl:copy>
+    </xsl:template>
+
+    <!-- Force quarkus-cassandra-client-test.version property for any artifactId matching cassandra-quarkus-integration-tests -->
+    <xsl:template match="/pom:project/pom:dependencies/pom:dependency/pom:version">
+        <xsl:copy>
+            <xsl:if test="starts-with(../pom:artifactId/text(), 'cassandra-quarkus-integration-tests')"><xsl:apply-templates select="@*" />${quarkus-cassandra-client-test.version}</xsl:if>
+        </xsl:copy>
```
aloubyansky, line 20: Is this still needed?

jamesnetherton: Yes, unless I'm misunderstanding the fundamentals of the platform generation.

The end goal is to be able to specify one version for the Cassandra artifacts that are listed in the BOM. And another for the test artifacts that we want to use. So this transform ensures that the tests artifacts are always set to the version that we specify in `quarkus-cassandra-client-test.version`.

aloubyansky: It'd be a bug if they aren't matching the version you configured in the `platformConfig`. Could you remove it and check whether it's really necessary?

jamesnetherton: Yes, you are right. I've removed the transform entirely.

Apologies for the confusion.

aloubyansky (approved):
Thanks @jamesnetherton!

PR #699, opened 2022-11-11
Remove the rhproduct profile

aloubyansky:


aloubyansky:
It moved to Red Hat internal repos.

PR #707, opened 2022-11-21
Workaround js-dsl fails on GraalVM 22.3/Quarkus 2.14.1

ppalaga:
….com/apache/camel-quarkus/issues/4285

### DIFF
--- a/generated-platform-project/quarkus-universe/properties/pom.xml
+++ b/generated-platform-project/quarkus-universe/properties/pom.xml
@@ -13,7 +13,7 @@
   <packaging>pom</packaging>
   <name>Quarkus Platform - Quarkus Universe - Quarkus Platform Properties</name>
   <properties>
-    <platform.quarkus.native.builder-image>graalvm</platform.quarkus.native.builder-image>
+    <platform.quarkus.native.builder-image>mandrel</platform.quarkus.native.builder-image>
   </properties>
   <dependencyManagement>
     <dependencies>
--- a/generated-platform-project/quarkus/properties/pom.xml
+++ b/generated-platform-project/quarkus/properties/pom.xml
@@ -12,7 +12,7 @@
   <packaging>pom</packaging>
   <name>Quarkus Platform - Core - Quarkus Platform Properties</name>
   <properties>
-    <platform.quarkus.native.builder-image>graalvm</platform.quarkus.native.builder-image>
+    <platform.quarkus.native.builder-image>mandrel</platform.quarkus.native.builder-image>
   </properties>
   <dependencyManagement>
     <dependencies>
--- a/pom.xml
+++ b/pom.xml
@@ -55,7 +55,7 @@
         <graalvmHome>${env.GRAALVM_HOME}</graalvmHome>
         <postgres.url>jdbc:postgresql:hibernate_orm_test</postgres.url>
 
-        <quarkus.version>2.14.0.Final</quarkus.version>
+        <quarkus.version>2.14.1.Final</quarkus.version>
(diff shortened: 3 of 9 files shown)


gsmet:
@ppalaga looks like the fix is not enough.

ppalaga:
> looks like the fix is not enough.

Amended. For some reason the Platform quarkus-bom was generated with graal-sdk 22.2.0. It is strange because the [core bom](https://repo1.maven.org/maven2/io/quarkus/quarkus-bom/2.14.1.Final/quarkus-bom-2.14.1.Final.pom)  manages 22.3.0. I have added a generator force entry for that too.

pom.xml
```diff
@@ -58,6 +58,11 @@
         <quarkus.version>2.14.0.Final</quarkus.version>
         <quarkus-bom.version>${quarkus.version}</quarkus-bom.version>
 
+        <!-- Workaround for https://github.com/apache/camel-quarkus/issues/4285
+             Remove once Camel Quarkus manages the same version of org.graalvm.js:js
+             as ${graalvm.version} in Quarkus -->
+        <graalvm.version>22.3.0</graalvm.version>
```
aloubyansky, line 64: This overrides the version from the `quarkus-bom`. We shouldn't be upgrading GraalVM version this way in the platform.

ppalaga: Hm... I wonder whether I should have sent this against the `main` branch @gsmet? 
Anyway, thanks @aloubyansky for pointing out that quarkus is at 2.14.0 in main. I thought it was 2.14.1

gsmet: Hmmm, no, it's my fault, I forgot to upgrade `main` to 2.14.1. Let me add a commit to this PR.

gsmet: I removed the workaround and upgraded to Quarkus 2.14.1.Final in a separate commit.

generated-platform-project/quarkus-universe/properties/pom.xml
```diff
@@ -13,7 +13,7 @@
   <packaging>pom</packaging>
   <name>Quarkus Platform - Quarkus Universe - Quarkus Platform Properties</name>
   <properties>
-    <platform.quarkus.native.builder-image>graalvm</platform.quarkus.native.builder-image>
+    <platform.quarkus.native.builder-image>mandrel</platform.quarkus.native.builder-image>
```
ppalaga, line 16: @gsmet should this not stay graalvm?

pom.xml
gsmet: I also created https://github.com/quarkusio/quarkus/pull/29405 as I think we always want to have the versions consistent and it's going to be a nightmare maintaining them separately.

generated-platform-project/quarkus/properties/pom.xml
```diff
@@ -12,7 +12,7 @@
   <packaging>pom</packaging>
   <name>Quarkus Platform - Core - Quarkus Platform Properties</name>
   <properties>
-    <platform.quarkus.native.builder-image>graalvm</platform.quarkus.native.builder-image>
+    <platform.quarkus.native.builder-image>mandrel</platform.quarkus.native.builder-image>
```
ppalaga, line 15: mandrel?

generated-platform-project/quarkus-universe/properties/pom.xml
gsmet: No we had to switch to use Mandrel by default because of a nasty bug in GraalVM that is not fixed yet and for which we don't have any ETA.

pom.xml
ppalaga: > I also created https://github.com/quarkusio/quarkus/pull/29405

Let me comment there

PR #714, opened 2022-11-29
DO NOT MERGE - Try upgrading sshd-common to 2.9.2

gsmet:


pom.xml
```diff
@@ -187,6 +187,8 @@
                                     <dependency>io.quarkiverse.vault:quarkus-vault:${quarkus-vault.version}</dependency>
                                     <dependency>io.quarkiverse.vault:quarkus-vault-deployment:${quarkus-vault.version}</dependency>
                                     <dependency>io.quarkiverse.vault:quarkus-vault-model:${quarkus-vault.version}</dependency>
+                                    <!-- Testing sshd-common upgrade -->
+                                    <dependency>org.apache.sshd:sshd-common:2.9.2</dependency>
```
ppalaga, line 191: It is vital to manage also sshd-core
```suggestion
                                    <dependency>org.apache.sshd:sshd-core:2.9.2</dependency>
                                    <dependency>org.apache.sshd:sshd-common:2.9.2</dependency>
```

PR #744, opened 2023-01-11
Upgrade to Quarkus 2.16.0.CR1

gsmet:


gsmet:
@radtriste  @rsynek not sure if you are aware of this issue with 2.16/main ^

radtriste:
@gsmet Thanks for the notification. This one is also spotted by our pipeline but did not have time to look into it yet.
Anyway we are targeting our Kogito 1.33 to be in Quarkus Platform 2.16 so we should correct this issue for the platform release.

PR #757, opened 2023-01-24
Update to Quarkus Qpid JMS 2.0.0.Alpha2

gemmellr:
Update to Quarkus Qpid JMS 2.0.0.Alpha2, uses Qpid JMS 2.2.0 against Quarkus 3.0.0.Alpha3

### DIFF
--- a/generated-platform-project/quarkus-qpid-jms/bom/pom.xml
+++ b/generated-platform-project/quarkus-qpid-jms/bom/pom.xml
@@ -61,17 +61,17 @@
       <dependency>
         <groupId>org.amqphub.quarkus</groupId>
         <artifactId>quarkus-qpid-jms-deployment</artifactId>
-        <version>2.0.0.Alpha1</version>
+        <version>2.0.0.Alpha2</version>
       </dependency>
       <dependency>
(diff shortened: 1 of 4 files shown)


gemmellr:
Closing as Quarkus main has a behaviour change that breaks native builds with the current extension, and so presumably this one will break when main becomes for 3.x at the weekend. I've done a new extension release for 2.x, I'll need to do another for 3.x also first.

gsmet:
@gemmellr what exactly did we break?

gemmellr:
Sorry, didnt see your message amongst a sea of others earlier.

The behaviour on main with 'missing' classes from e.g optional dependencies looks to have changed, probably following a recent change in how classes registered for reflection are passed to graal. I say probably since that change had various related PRs (which I used for inspiration on what changed and how to fix things) adjusting other extensions etc to only optionally register certain things for reflection, in order to remove warnings the registration related change introduced. In the quarkus-qpid-jms case, the 'missing' class was not the direct one being registered but rather other indirect classes used by the class being registered, and that seems to make the compilation fail now. I adjusted the extension to look for an optional dep class before registering the intermediate for reflection, and things went back to working as before.

PR #763, opened 2023-02-08
Upgrade to Quarkus 2.16.2.Final

gsmet:


gemmellr:
Do you think the reflection registration changes from Quarkus main would ever be backported to 2.16.x? Should we backport #761 in case?

gsmet:
I don't think we will backport anything breaking but let's be on the safe side. I will backport it once this is in (I don't want to interrupt CI or I won't be able to release today)

gemmellr:
Great, thanks.

gsmet:
Well... you got lucky... :)

gsmet:
Well, the CI build failed so I had to restart it anyway. So I also pushed your changes.

PR #785, opened 2023-03-09
Update CI to Java 17

gsmet:
This is done in preparation of the inclusion of Camel Quarkus 3.0.0-M1.

It might be temporary if Camel goes back to Java 11.

/cc @ppalaga

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -47,7 +47,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        java: [ 11 ]
+        java: [ 17 ]
     steps:
       - uses: actions/checkout@v3
       - name: Get Date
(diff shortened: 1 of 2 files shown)


.github/workflows/build.yml
```diff
@@ -47,7 +47,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        java: [ 11 ]
```
aloubyansky, line 50: Perhaps we need both?

zbendhiba:
Thanks 👍🏾

.github/workflows/build.yml
aloubyansky: Either we could kindly ask the CEQ team to annotate their tests with supported Java versions or will need some external config.

PR #815, opened 2023-04-14
CEQ: move test pattern config from the XSL template to the POM config

aloubyansky:
@ppalaga @zbendhiba would you mind moving the test pattern config from the XSLT to the POM?


zbendhiba:
LGTM!

ppalaga (approved):
Looks good, thanks!

zbendhiba (approved):

PR #816, opened 2023-04-18
Upgrade to Camel Quarkus 2.13.3

aldettinger:
Here is a Camel update for Quarkus Platform 2.13.

There is an issue when testing locally, I suspect it should work on ci. So setting this pr as draft for now.


aldettinger:
Ok, after investigation it was only flaky test and tests failing before this pr. Let's turn this pr ready for review.

gsmet:
If these tests are known to fail, please add it to the ignored tests.

jamesnetherton:
I think this change may be the cause of the Kogito test failures:

https://github.com/apache/camel-quarkus/blob/2.13.x/poms/bom/src/main/generated/flattened-full-pom.xml#L19996-L20001

The tests are using WireMock which depends on `jetty-server`.

Is there a way we could add `servlet-api` into `kogito-quarkus-serverless-workflow-integration-test`?

aloubyansky:
Would Kogito and Camel still be compatible when used in the same application?

jamesnetherton:
`WireMock` is a test dependency, so would not be on the runtime classpath. If anything else in Kogito uses `jetty-server`, then I guess it'd be a case of adding `servlet-api` to the app depedencies (unless some other extension is already adding it).

aloubyansky:
@radtriste could you confirm the `servlet-api` exclusion from the `jetty-server` isn't breaking Kogito extensions? Thanks.

radtriste:
@aloubyansky the jetty dependency on the integration-test is only for a test dependency.
The kogito-serverless-workflow extension should be working fine.
@spolti or @cristianonicolai could confirm that.

aloubyansky:
Ok, so `quarkus-camel-bom` doesn't affect Kogito extensions but it does affect WireMock users. That doesn't seem insignificant. A way to avoid it would be to exclude `javax.servlet-api` from those CEQ extensions that have it among their transitive dependencies. Would that make sense @jamesnetherton ?

jamesnetherton:
Yes it does make sense. WDYT @ppalaga?

Also, I guess there's no way to resolve this in the platform? I.e we'd need another CQ release?

aloubyansky:
I think we/you could re-configure it in the platform. You'd need to know which extensions to adjust.

ppalaga:
Reconfiguring in the Platform sounds good. 

The extensions depending on jetty-server are

* camel-quarkus-cometd
* camel-quarkus-hdfs

Both are JVM-only, so they are not a part of the platform at all.

jamesnetherton:
So how is the reconfiguring done? Are there options available on the bom-generator plugin? Or does it need to be done via an XSLT transform?

aloubyansky:
@jamesnetherton you could do it using `dependencyManagement` config in the member configuration, here is an example:
```
diff --git a/generated-platform-project/quarkus-camel/bom/pom.xml b/generated-platform-project/quarkus-camel/bom/pom.xml
index 8848d89ac..d8a38e281 100644
--- a/generated-platform-project/quarkus-camel/bom/pom.xml
+++ b/generated-platform-project/quarkus-camel/bom/pom.xml
@@ -1547,6 +1547,12 @@
         <groupId>org.apache.camel.quarkus</groupId>
         <artifactId>camel-quarkus-core</artifactId>
         <version>2.13.3</version>
+        <exclusions>
+          <exclusion>
+            <groupId>javax.servlet</groupId>
+            <artifactId>javax.servlet-api</artifactId>
+          </exclusion>
+        </exclusions>
       </dependency>
       <dependency>
         <groupId>org.apache.camel.quarkus</groupId>
@@ -7803,12 +7809,6 @@
         <groupId>org.eclipse.jetty</groupId>
         <artifactId>jetty-server</artifactId>
         <version>9.4.51.v20230217</version>
-        <exclusions>
-          <exclusion>
-            <groupId>javax.servlet</groupId>
-            <artifactId>javax.servlet-api</artifactId>
-          </exclusion>
-        </exclusions>
       </dependency>
       <dependency>
         <groupId>org.eclipse.jetty</groupId>
diff --git a/generated-platform-project/quarkus-universe/bom/pom.xml b/generated-platform-project/quarkus-universe/bom/pom.xml
index e88a76874..419151890 100644
--- a/generated-platform-project/quarkus-universe/bom/pom.xml
+++ b/generated-platform-project/quarkus-universe/bom/pom.xml
@@ -12259,6 +12259,12 @@
         <groupId>org.apache.camel.quarkus</groupId>
         <artifactId>camel-quarkus-core</artifactId>
         <version>2.13.3</version>
+        <exclusions>
+          <exclusion>
+            <groupId>javax.servlet</groupId>
+            <artifactId>javax.servlet-api</artifactId>
+          </exclusion>
+        </exclusions>
       </dependency>
       <dependency>
         <groupId>org.apache.camel.quarkus</groupId>
@@ -19757,12 +19763,6 @@
         <groupId>org.eclipse.jetty</groupId>
         <artifactId>jetty-server</artifactId>
         <version>9.4.51.v20230217</version>
-        <exclusions>
-          <exclusion>
-            <groupId>javax.servlet</groupId>
-            <artifactId>javax.servlet-api</artifactId>
-          </exclusion>
-        </exclusions>
       </dependency>
       <dependency>
         <groupId>org.eclipse.jetty</groupId>
diff --git a/pom.xml b/pom.xml
index 517cb598b..db76ba5ab 100644
--- a/pom.xml
+++ b/pom.xml
@@ -418,6 +418,15 @@
                                         <lastDetectedBomUpdate>io.quarkus.platform:quarkus-camel-bom:2.13.7.Final</lastDetectedBomUpdate>
                                         <next>${platform.groupId}:quarkus-camel-bom:${platform.version}</next>
                                     </release>
+                                    <dependencyManagement>
+                                        <dependency>org.eclipse.jetty:jetty-server:9.4.51.v20230217</dependency>
+                                        <dependencySpec>
+                                            <artifact>org.apache.camel.quarkus:camel-quarkus-core:${camel-quarkus.version}</artifact>
+                                            <exclusions>
+                                                <exclusion>javax.servlet:javax.servlet-api</exclusion>
+                                            </exclusions>
+                                        </dependencySpec>
+                                    </dependencyManagement>
```

aloubyansky:
> The extensions depending on jetty-server are
> 
> * camel-quarkus-cometd
> * camel-quarkus-hdfs
> 
> Both are JVM-only, so they are not a part of the platform at all.

To clarify, they are in the BOM but unlisted in the descriptor.

jamesnetherton:
> To clarify, they are in the BOM but unlisted in the descriptor.

Yes, correct.

jamesnetherton:
Turns out that `cometd` and `hdfs` already have exclusions for `servlet-api`. So it should just be a case of removing the exclusion from `jetty-server`.

I'll close & recreate this PR with that change.

jamesnetherton:
https://github.com/quarkusio/quarkus-platform/pull/818

PR #820, opened 2023-04-21
Upgrade Cassandra to 1.2.0

aloubyansky:
FYI @absurdfarce


absurdfarce:
Many thanks for this @aloubyansky!  This looks quite similar to some of Alexandre's older platform updates and seems to contain precisely the updates I'd expect given our conversation.  With that in mind... is there anything _else_ that needs to be done to make sure we're ready for the upcoming platform release?

aloubyansky:
No, that's it, Cassandra is included.

PR #822, opened 2023-04-25
Upgrade to Quarkus 3.0.1.Final

gsmet:

PR #824, opened 2023-04-26
Upgrade Quarkus to 3.0.1

aloubyansky:
Also upgrade the platform generator to 0.0.88 that filters out test scoped dependencies.

Supersedes https://github.com/quarkusio/quarkus-platform/pull/822

### DIFF
--- a/generated-platform-project/quarkus-universe/bom/pom.xml
+++ b/generated-platform-project/quarkus-universe/bom/pom.xml
@@ -4928,17 +4928,17 @@
       <dependency>
         <groupId>io.apicurio</groupId>
         <artifactId>apicurio-common-rest-client-vertx</artifactId>
-        <version>0.1.13.Final</version>
+        <version>0.1.15.Final</version>
       </dependency>
       <dependency>
         <groupId>io.apicurio</groupId>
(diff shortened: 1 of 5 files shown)


aloubyansky:
Ha, here you go @metacosm
```
java.lang.RuntimeException: io.quarkus.builder.BuildException: Build failure: Build failed due to errors
	[error]: Build step io.quarkiverse.operatorsdk.deployment.OperatorSDKProcessor#createConfigurationServiceAndOperator threw an exception: java.lang.RuntimeException: Incompatible Quarkus version found: "3.0.1.Final", expected: "3.0.0.Final"
```
So, we disable JOSDK after all?

maxandersen:
that version check cannot be right. yeah; operator SDK will have to be removed from platform until it does not hard fail on even the slightest dependency update.

aloubyansky:
To potentially save @gsmet's time, here is a PR upgrading to 3.0.1 and disabling JOSDK https://github.com/quarkusio/quarkus-platform/pull/825

gsmet:
Yes, I won't have the time to wait for another release of the JOSDK so let's drop it...

@metacosm we release Quarkus every week, I don't see how adding this check could be a good idea, we need to drop it.

PR #826, opened 2023-04-26
Re-include QOSDK in version 6.0.2

metacosm:

Thanks!


aloubyansky:
@metacosm will 6.0.1 fail with Quarkus 3.0.2.Final?

aloubyansky (changes requested):
Let's clarify the version update strategy for the JOSDK

metacosm:
> @metacosm will 6.0.1 fail with Quarkus 3.0.2.Final?

No, the extension defaults to outputting a warning now, previously the default was to fail the build.

aloubyansky (approved):

PR #831, opened 2023-05-04
Define system property for optaplanner test

rsynek:
Fixes the devui integration test for OptaPlanner; required by https://github.com/quarkusio/quarkus-platform/pull/829.


aloubyansky (approved):
We should probably recognize and configure original plugin system properties.

PR #879, opened 2023-07-01
Fix Qpid test version

aloubyansky:

Thanks!


gemmellr:
Did you just update this manually? Should I be doing the same? I didnt for my PR as that seemed like it would break the PR checks the generated project is correct.

Ah, no...I see there was since an overlapping PR merged for updating the amazon bits, opened just after I opened my PR for quarkus-qpid-jms, which presumably means it fixed the 'uses wrong prop name for same resolved  value' issue after my PR was raised since it changed the version of the amazon bits, so it differed from the quarkus-qpid-jms version again and the generator went back to using the correct quarkus-qpid-jms prop name after that later PR was merged (with it still using the wrong value from when my PR was merged).

Answered my own question, but I'm just going to post this comment since I've typed it :)

aloubyansky:
Yes, it's a known issue when there are multiple properties sharing the same value. I need to think about a better approach.

gemmellr:
Yep, I recalled it coming up before which is why I didnt bother to mention it for my PR, just got curious when I saw the fix commit.

PR #867, opened 2023-06-21
[2.16.x] Disable Camel SJMS and LRA tests

jamesnetherton:
Relates to https://github.com/quarkusio/quarkus-platform/pull/864#issuecomment-1601007278.

PR #880, opened 2023-07-03
[2.16.x] Disable Camel JMS and JTA tests

jamesnetherton:
Follow up on #867. Looks like I missed a few tests that have been causing issues for other PRs.


gsmet:
Let's keep this PR open for now but it seems the problem was due to a disk space issue, that we solved.

We'll see how it goes for the next runs.

PR #922, opened 2023-08-09
Revert Downgrade angus-mail to 1.1.0

ppalaga:
fix https://github.com/quarkiverse/quarkus-cxf/issues/899


ppalaga:
This may not be a regression, because there is (strangely) no change in the generated BOMs.

aloubyansky:
It's because the CEQ BOM includes 2.0.2 version of it and it wins during alignment.

ppalaga:
Is this good to merge in spite of the CI failures?

gsmet (approved):
Yes, we can merge.

Both failing components need a new release to be compatible with 3.3.

PR #938, opened 2023-08-24
Upgrade to Google Cloud Services 2.4.0

loicmathieu:


loicmathieu:
Test failure seems unrelated:
```
Caused by: java.lang.IllegalStateException: Cannot set quarkus.http.redirect-insecure-requests without enabling SSL.
```

aloubyansky:
@loicmathieu could you run `./mvnw -Dsync` and commit the changes though, please?

loicmathieu:
I did it but not include all changes because, for an unknown reason, it changes unrelated version (version of Quarkus extensions!).
I double check if I miss something.

loicmathieu:
@aloubyansky I added a commit to fix some issues with CXF dependency using a fixed version instead of a property.

Maven also updated the `quarkus-qpid-jms-integration-test` pom to use a fixed version instead of a property, I didn't commit it as it seems a bug to me.

aloubyansky:
That's ok, please, always include all the changes, otherwise, it'll complain.

The generator prefers properties instead of specific values, it looks like both CXF and GCS extensions had the same versions but used different properties, now that you upgraded your extensions to 2.4.0 it became unambiguous and it replaced the CXF 2.3.0 with its property.

loicmathieu:
OK, I'll added the Qpid changes then

aloubyansky (approved):

gsmet:
I fixed the issue and force pushed.

PR #946, opened 2023-08-28
Upgrade to Quarkus 3.3.1

gsmet:

Thanks!


gsmet:
@ppalaga @zbendhiba @jamesnetherton an important thing is that we haven't updated Infinispan in 3.2 and I'm not sure we will. Not sure if it's possible to support both at the same time. If not, we will need to have a discussion.

zbendhiba:
Thanks @gsmet. I'll look at it this, as we are about to upgrade our main to Quarkus 3.3

PR #948, opened 2023-09-01
Upgrade to Camel Quarkus 3.2.0

zbendhiba:

Thanks!


aloubyansky:
@zbendhiba thanks for noticing, it looks like it's been going on for quite some time. Here is a PR that will let Debezium control its version https://github.com/quarkusio/quarkus-platform/pull/949

aloubyansky:
We should upgrade Debezium to 2.3.0 too, I suppose, since it's available.

aloubyansky:
@zbendhiba if you rebase now, it should preserve Debezium 2.3.0. I think you could also update it to 2.3.2 in a separate commit. Thanks!

aloubyansky:
@zbendhiba so CEQ 3.2.0 can't be added to 3.3 and main yet?

zbendhiba:
@aloubyansky  We can push to main and 3.3.  There will be some tests failing, but it's already the case in 3.3 branch

aloubyansky:
Will it introduce new regressions? If not, I'd proceed with the upgrade.

zbendhiba:
@aloubyansky I missed your last comment. I think we can at least put this final in 3.3. If we can catch a quarkus platform release next week, I can plan releasing one version of camel-quarkus with 3.3

aloubyansky:
Sure, go ahead @zbendhiba

PR #1007, opened 2023-10-18
Upgrade to Quarkus 3.5.0

gsmet:


gsmet:
@ppalaga @zbendhiba see the Camel failures above (but they are probably the same we had with the CR1)

@vsevel see the Vault failures above (but they are probably the same we had with the CR1)

vsevel:
Can you check @kdubb I am away from my desk. Thanks.

kdubb:
@vsevel On it

kdubb:
@vsevel These are the same errors we looked into with #182. The string version of the `ASN1Util` is producing the old behavior. So there must be a misalignment of BouncyCastle versions.

kdubb:
@gsmet Actually this looks like https://github.com/quarkiverse/quarkus-vault/pull/182 isn't included in this build. How can we tell if it is still using the `.CR1` release of Vault?

kdubb:
@vsevel Just to tie things up. This failed because it was still using `3.2.0` of `quarks-vault`.

PR #1019, opened 2023-11-04
[3.2] Update QOSDK to 6.3.3

metacosm:

Thanks!

### DIFF
--- a/generated-platform-project/quarkus-operator-sdk/bom/pom.xml
+++ b/generated-platform-project/quarkus-operator-sdk/bom/pom.xml
@@ -61,42 +61,47 @@
       <dependency>
         <groupId>io.javaoperatorsdk</groupId>
         <artifactId>micrometer-support</artifactId>
-        <version>4.4.2</version>
+        <version>4.4.4</version>
       </dependency>
       <dependency>
         <groupId>io.javaoperatorsdk</groupId>
         <artifactId>operator-framework-core</artifactId>
-        <version>4.4.2</version>
+        <version>4.4.4</version>
+      </dependency>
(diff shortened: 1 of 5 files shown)


metacosm:
Not sure what this needs, I did run `./mvnw -Dsync` before opening the PR (and did so again, just now, to double check), and there isn't any more changes to commit.

aloubyansky:
It's because you have the Gradle Maven repo enabled in your settings while it's not enabled in the CI and most of our local configs, it results in gradle-tooling-api added to the quarkus-bom.

metacosm:
Hmm, I don't recall doing anything specific. I'm not even sure what this "Gradle Maven repo" is supposed to refer to, to be honest… What do I need to do to fix the issue?

aloubyansky:
It seems like you have https://repo.gradle.org/artifactory/libs-releases or a mirror of it enabled in your settings. If you disable it it should fix the check.

metacosm:
I don't, though. `cat ~/.m2/settings.xml | grep gradle` returns nothing. Or if I do, I have no idea where it is and I certainly didn't set it up.

metacosm:
Of note, I've opened several similar PRs with the exact same configuration without issue before so I don't know why this would failing now.

aloubyansky:
And if you re-sync it it's still the same?

metacosm:
> And if you re-sync it it's still the same?

yes

metacosm:
Hmm, that said, the PR was opened by a bot, so it probably has different settings than I do. That said, this also isn't the first PR opened by the bot either so 🤷🏼

PR #1040, opened 2023-11-22
[3.2] Upgrade to Quarkus CXF 2.2.6

ppalaga:


ppalaga:
This may fail, because Quarkus CXF 2.2.6 is not in Central yet

ppalaga:
I have pushed some dummy dummy change to re-trigger the job now that all QCXF artifacts are available in Central.

PR #1052, opened 2023-11-27
Unlist quarkus-resteasy-qute and quarkus-resteasy-reactive-qute

ia3andy:
part of #37000

### DIFF
--- a/src/main/resources/extensions-overrides.json
+++ b/src/main/resources/extensions-overrides.json
@@ -52,6 +52,20 @@
             "metadata": {
                "unlisted": true
             }
+        },
+        {
+            "group-id": "io.quarkus",
+            "artifact-id": "quarkus-resteasy-qute",
+            "metadata": {
+               "unlisted": true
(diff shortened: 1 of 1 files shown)


ia3andy:
cc @mkouba @gsmet

mkouba (approved):
Looks good.

aloubyansky:
Was there a PR unlisting these in core?

ia3andy:
@aloubyansky should we unlist them in core? if that's the case we don't need to do it here...

aloubyansky:
Yes, they should be unlisted in core unless there is a reason not to.

ia3andy:
@aloubyansky ah but they should be also unlisted here (just for this next platform release as it's not yet in core)

aloubyansky:
Sure, no problem.

gsmet (approved):
OK, let's merge it for now and, ideally, let's drop it when fixed in core.

PR #1062, opened 2023-12-12
Upgrade to Quarkus 3.6.2

gsmet:


gsmet:
@ppalaga @zbendhiba @jamesnetherton looks like we have an issue with the latest Infinispan micro.

PR #1064, opened 2023-12-13
Upgrade to Debezium 2.4.2.Final

Naros:
@gsmet would it be possible to cut a platform release with this?


gsmet:
We can include it in next week release.

@aloubyansky please let it open, I will merge it and backport it right away so that we don't forget.

PR #1074, opened 2023-12-20
Avoid asking for pin for gpg signing on CI

gsmet:

Thanks!


maxandersen:
am i missing something or couldn't that profile  not just be in the shared parent pom ?

PR #1085, opened 2024-01-15
Upgrade to Debezium 2.5.0.Final

Naros:
Upgrade Platform to Debezium 2.5.0.Final baseline

- [x] Address Hibernate ORM issue in Test Suite, per request from Guillaume

The above should be fixed with the Debezium upgrade to 2.4+, which uses OpenTelemetry over OpenTracing, @gsmet.


gsmet (approved):
Cool, thanks!

PR #1114, opened 2024-02-14
Upgrade to Quarkus 3.8.0.CR1

gsmet:

Thanks!

### DIFF
--- a/generated-platform-project/quarkus-camel/bom/pom.xml
+++ b/generated-platform-project/quarkus-camel/bom/pom.xml
@@ -902,7 +902,7 @@
       <dependency>
         <groupId>io.smallrye.reactive</groupId>
         <artifactId>smallrye-reactive-messaging-camel</artifactId>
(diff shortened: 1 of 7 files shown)


gsmet:
@ppalaga @zbendhiba @jamesnetherton Not sure about all the JMS failures but I wonder if they could be related to us updating the Dev Services image in https://github.com/quarkusio/quarkus/pull/38638/files to fix a CI issue we were having (it was a micro bump).

Note that 3.7.3 also fails with the exact same failures.

jamesnetherton:
Our tests use whatever quarkus-artemis dev services provides (activemq-artemis-broker:artemis.2.31.0).

Those failures are probably related to lack of disk space. Not sure what else to do other than disable those tests.

I actually had to deal with something similar here https://github.com/apache/camel-quarkus/pull/5726. Not sure why space suddenly became an issue because builds were running fine until that point...

PR #1129, opened 2024-02-26
Upgrade to Camel Quarkus 3.8.0

aldettinger:


pom.xml
```diff
@@ -602,6 +602,7 @@
                                     <dependency>log4j:log4j</dependency>
                                     <!-- AssertJ -->
                                     <dependency>org.assertj:assertj-core</dependency>
+                                    <dependency>org.apache.camel.quarkus:camel-quarkus-beanio</dependency>
```
aldettinger, line 605: The camel-quarkus-beanio extension seems to have incomplete metadata. At this stage, the plan would be to add it back in a future release once the root cause has been identified.

aldettinger:
@gsmet After working around the Camel Quarkus failing tests, there are a still a few quarkiverse tests failing:
 * io.quarkiverse.googlecloudservices.it.PubSubResourceTest.testPubSub
 * io.quarkiverse.operatorsdk.it.EmptyOperatorLevelNamespacesTest.reconcilersWithoutSpecificNamespacesShouldWatchAllNamespaces
 * io.quarkiverse.operatorsdk.it.OperatorSDKResourceTest
 * io.quarkiverse.operatorsdk.it.SetOperatorLevelNamespacesTest.dependentAnnotationsShouldAppearInConfiguration

Are these known issues ? Does it prevent from merging ?

jamesnetherton:
> After working around the Camel Quarkus failing tests, there are still a few quarkiverse tests failing:

@aldettinger I don't have karma to push to your branch. This may help things https://github.com/aldettinger/quarkus-platform/pull/1.

gsmet:
@loicmathieu the Camel upgrade is causing upgrades to the Google libraries (see the Google BOM in the PR) and apparently it's causing some test issues. Could you have a look?

gsmet:
I rebased to drop the merge commits. Let's see how CI goes with your latest changes.

metacosm:
> @metacosm do the Operator SDK failures above ring a bell?

Difficult to tell without more details but QOSDK (and more generally, Fabric8 client) won't work properly with Quarkus 3.7.4 and above in native because of the commons-compress update. We have workarounds but we're hoping for a 1.26.1 release of commons-compress that should address the issue.

loicmathieu:
Latest version of the Google Cloud Library BOM didn't work in native, see https://github.com/quarkiverse/quarkus-google-cloud-services/actions/runs/7999885847/job/21848410829?pr=582

It may be better if Quarkus Camel would not include directly the Google Library BOM but uses the version from the Quarkus Google Cloud Services to avoid dependency divergence (this is what is done for ex for Micrometer Stackdriver). Would it be possible @jamesnetherton ?

This is not the first time that a new version of the Google Cloud Library BOM didn't work in native, I will try to get in touche with them to see how we can work together to smooth things up.

jamesnetherton:
> Would it be possible @jamesnetherton ?

We could try to experiment with that. I'm not sure of the best way to handle it...

Camel is eagerly updating to the latest Google Library BOM via dependabot. If I have to force downgrade and override dependencies in Camel Quarkus, it could potentially break the Camel Google components. So I'd have to downgrade the BOM version in Camel itself.

Thus we could end up in a position where neither CQ or Camel can upgrade until Quarkus Google Cloud Services has upgraded. 

Maybe if we did more frequent releases of Quarkus Google Cloud Services it'd help a bit?

> new version of the Google Cloud Library BOM didn't work in native

Do you remember which issue(s)? The only one I'm aware of was with BigQuery and Apache Arrow Netty being incompatible with Quarkus Netty. That is resolved in the latest releases.

aloubyansky:
@loicmathieu in case it helps, you could configure groupIds of artifacts you would like to control with `ownGroupIds` element, e.g. https://github.com/quarkusio/quarkus-platform/blob/main/pom.xml#L465-L467
For now there is nothing configured for the GCS member, which means it "owns" only the groupId of its quarkiverse BOM.

loicmathieu:
@jamesnetherton well, if it's Camel and not Camel Quarkus that upgrades the dependency, it should stay as is.

> new version of the Google Cloud Library BOM didn't work in native

It's not the first time, and with the BigQuery/Arrow issue it causes crash multiple time as we had to create a substitution then remove it ...

@aloubyansky seems interested, I'll put this in my TODO list.

To unblock the issue the BOM should be downgraded to 26.29.0 or the issue shown here should be fixed: https://github.com/quarkiverse/quarkus-google-cloud-services/pull/582

pom.xml
aldettinger: Many thanks to @jamesnetherton who found the root [cause](https://github.com/apache/camel-quarkus/issues/5803). It's probably time to add a comment pointing to this issue + squashing commits.

gsmet (approved):

gsmet:
Merging this but we will probably need to sort out the Google Cloud libraries situation. If Camel wants to use the latest version, they will need to contribute to the Google Cloud libraries extensions to make sure they actually work in native.

PR #1130, opened 2024-02-26
Vault 3.5.0

vsevel:
https://github.com/quarkiverse/quarkus-vault/releases/tag/3.5.0 - Quarkus 3.8, JDK 17 + bug fix
note that the synch operation modified also assertj version to `3.25.3` (like in vault), which I did not include in this PR

### DIFF
--- a/generated-platform-project/quarkus-universe/bom/pom.xml
+++ b/generated-platform-project/quarkus-universe/bom/pom.xml
@@ -9378,17 +9378,17 @@
       <dependency>
         <groupId>io.quarkiverse.vault</groupId>
         <artifactId>quarkus-vault-deployment</artifactId>
-        <version>3.4.0</version>
+        <version>3.5.0</version>
       </dependency>
       <dependency>
         <groupId>io.quarkiverse.vault</groupId>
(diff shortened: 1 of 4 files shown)


vsevel:
if I ru the sync, it adds:
```
     <dependency>
       <groupId>org.assertj</groupId>
       <artifactId>assertj-core</artifactId>
-      <version>3.24.2</version>
+      <version>3.25.3</version>
       <scope>test</scope>
     </dependency>
```
what should I do?

aloubyansky:
@vsevel all the changes are supposed to be added to the commit, otherwise it will fail the check. Then it'll be up to us to review the changes before merging.
If you have concerns about some of them, please add them all and add your comments for us to review. Thanks.

vsevel:
It seemed strange to me that a bump in a test module of an ext would lead to a version change in the plateform ...

aloubyansky:
It's a bump in the corresponding test module in the platform though. So it is consistent with the original test dependencies.

gsmet (approved):

gsmet:
Thanks!

PR #1172, opened 2024-03-27
Quarkus Vault 4.0.0

vsevel:
still time for `3.9` @gsmet ?
cc @kdubb


gsmet:
I confirm it's going to be part of 3.9 if all goes well!

PR #1182, opened 2024-04-16
Upgrade to Quarkus 3.10.0.CR1

gsmet:

Thanks!


gsmet:
@ppalaga there are several issues with CXF and Camel. I think part of them are due to https://github.com/quarkusio/quarkus/pull/39295 . Maybe we should produce a build item with the packaging rather than using the config that is not really stable.
But it's probably something for the future though.

/cc @dmlloyd

ppalaga:
> there are several issues with CXF and Camel. I think part of them are due to [quarkusio/quarkus#39295](https://github.com/quarkusio/quarkus/pull/39295)

Yeah, I have a fix ready in Quarkus CXF. Let me release.

PR #1213, opened 2024-05-27
Upgrade to Camel Quarkus 3.11.0

zbendhiba:

Thanks!


zbendhiba:
Hello @aloubyansky @gsmet 
I noticed that the `./mvnw -Dsync`moved many `${camel-quarkus.version}`to `3.11.0`. I'm not sure if that is expected or not ?

aloubyansky:
It's ok, it's a coincidence of having Quarkus core version matching Camel Quarkus version and the generator could not decide which property to use.

gsmet (approved):

gsmet:
> I noticed that the ./mvnw -Dsyncmoved many ${camel-quarkus.version}to 3.11.0. I'm not sure if that is expected or not ?

Yeah it is a bit annoying as it generates a lot of noise in the PR but basically Maven randomly (probably not randomly but you get the point) chooses a property that match the version. So when updating a given version, and you have a property that matches, it will pick one and it might not be the previous one.

zbendhiba:
Thanks for the explanations @gsmet @aloubyansky

PR #1215, opened 2024-05-28
Upgrade to Quarkus CXF 3.8.4

JiriOndrusek:

Thanks!

### DIFF
--- a/generated-platform-project/quarkus-camel/integration-tests/camel-quarkus-integration-test-cxf-soap-grouped/pom.xml
+++ b/generated-platform-project/quarkus-camel/integration-tests/camel-quarkus-integration-test-cxf-soap-grouped/pom.xml
@@ -73,7 +73,7 @@
     <dependency>
       <groupId>io.quarkiverse.cxf</groupId>
       <artifactId>quarkus-cxf-test-util</artifactId>
-      <version>${quarkus-cxf.version}</version>
+      <version>3.8.3</version>
       <scope>test</scope>
     </dependency>
     <dependency>
@@ -124,6 +124,9 @@
           <dependenciesToScan>
             <dependency>org.apache.camel.quarkus:camel-quarkus-integration-test-cxf-soap-grouped</dependency>
           </dependenciesToScan>
+          <excludes>
+            <exclude>**/CxfWsTrustCxfWayTest*</exclude>
+          </excludes>
         </configuration>
       </plugin>
     </plugins>
@@ -150,6 +153,9 @@
                   <systemProperties>
                     <native.image.path>${project.build.directory}/${project.build.finalName}-runner</native.image.path>
                   </systemProperties>
+                  <excludes>
+                    <exclude>**/CxfWsTrustCxfWayTest*</exclude>
+                  </excludes>
                 </configuration>
               </execution>
             </executions>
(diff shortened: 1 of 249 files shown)


JiriOndrusek:
The issue during the build is fixed in camel-quarkus 3.8.4

gsmet:
Do we have to wait for Camel Quarkus 3.8.4 or should we merge this?

The plan is to release Quarkus 3.8.5 next Wednesday (tentatively) so it needs to be decided soon.

Thanks!

JiriOndrusek:
I was wrong with my statement that the fix is present in Camel Quarkus 3.8.4. Issue is test only related. As a workaround I disabled the failing test in this PR and reported the [ticket](https://github.com/apache/camel-quarkus/issues/6128) to enable the test with the new camel quarkus version.

JiriOndrusek:
Build in the failed CI build is **successful**. (the disablement of the test generated a change in the pom.xml, which made the CI job failed)

generated-platform-project/quarkus-camel/integration-tests/camel-quarkus-integration-test-cxf-soap-grouped/pom.xml
```diff
@@ -73,7 +73,7 @@
     <dependency>
       <groupId>io.quarkiverse.cxf</groupId>
       <artifactId>quarkus-cxf-test-util</artifactId>
-      <version>${quarkus-cxf.version}</version>
+      <version>3.8.3</version>
```
ppalaga, line 76: @gsmet pointed me to this. Per se, it is not correct but not that important, because it is a test dependency in a test project. The main thing is, that the BOMs are correct. Hard to say, whether it is a bug in the generator plugin. In this situation, where quarkus-cxf-test-util is not managed and the version of all qcxf artifacts in `org.apache.camel.quarkus:camel-quarkus-bom` is 3.8.3, the generator can't perhaps do any other guess. WDYT, @aloubyansky? The bottom line is that both QCXF and CQ have their own test BOMs which we ideally might want to align and use in the Platform, but it is perhaps not worth of the effort, to solve such small details as this one.

aloubyansky: If a test dependency isn't managed by platform BOMs then its version is copied from the original test POM for that test.

ppalaga: I see. Then it works as designed. What is your take on aligning and using the test BOMs? It is not worth the trouble, is it?

aloubyansky:
```
test BOMs which we ideally might want to align and use in the Platform
```
could you please clarify what you mean specifically?

ppalaga: Here are the two test boms: 
https://github.com/quarkiverse/quarkus-cxf/blob/main/bom-test/pom.xml#L21-L25
https://github.com/apache/camel-quarkus/blob/main/poms/bom-test/pom.xml#L65-L71

The CQ test BOM imports the QCXF test BOM so to avoid this kind of issues, we'd need to not only use the BOMs in the generated tests in the Platform, but we also would have to align the two BOMs so that newer QCXF gets propagated to the CQ test BOM. As I said this is perhaps too much work for too little benefit.

PR #1230, opened 2024-06-27
Upgrade platform generator to 0.0.107

aloubyansky:
It includes fixes related to manifesting, so not really critical for building and testing use-cases.


aloubyansky:
Interesting, with Maven 3.9.8 it uses a property for the nexus plugin version instead of the actual value. The Maven wrapper is at 3.9.6 currently. We should probably upgrade it as a follow up PR.

gsmet (approved):

PR #1262, opened 2024-08-27
Upgrade Qute Web to 3.1.0

ia3andy:


ia3andy:
@gsmet it seems the pdfbox was using a wrong version from properties right?

mkouba (approved):

gsmet:
Yeah, that's just the bom generator that does that. It takes one of the property that matches the version.

ia3andy:
@gsmet error doesn't seem related, we need this for 3.14 platform release (if error is unrelated)

ia3andy:
I rebased if fix was applied on main

ia3andy:
ok I see you merged PR with the same error, I am merging this one

PR #1289, opened 2024-09-20
Upgrade to Google Cloud extension pack 2.12.1

loicmathieu:


aloubyansky (approved):
Thanks @loicmathieu 
And I figured out the plugin version diff and the sync check failure.
It depends on the Maven version. The check is using the wrapper but you are using a different version and that results in this diff.
But I have a fix for it in the generator.

loicmathieu:
Oh! Sorry about that @aloubyansky I never use the wrapper and my Maven may not be of the last version, I'll take care next time.

PR #1312, opened 2024-10-04
Upgrade to the platform generator 0.0.116

aloubyansky:
This PR includes the following changes:

- Upgrade the platform generator to 0.0.116;
- change deprecated systemProperties to systemPropertyVariables in the failsafe config;
- includes an integration test for the platform Maven plugin that creates a sample Quarkus application project and builds it.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -84,7 +84,7 @@
         <quarkus-vault.version>4.1.0</quarkus-vault.version>
         <quarkus-operator-sdk.version>6.8.1</quarkus-operator-sdk.version>
 
-        <quarkus-platform-bom-generator.version>0.0.113</quarkus-platform-bom-generator.version>
+        <quarkus-platform-bom-generator.version>0.0.116</quarkus-platform-bom-generator.version>
         <build-helper-maven-plugin.version>3.1.0</build-helper-maven-plugin.version>
         <maven-plugin-plugin.version>3.6.1</maven-plugin-plugin.version>
         <version.surefire.plugin>3.0.0</version.surefire.plugin>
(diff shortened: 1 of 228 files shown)


gsmet (approved):

pom.xml
```diff
@@ -631,6 +632,26 @@
...
+                                             to run both the unit and the integration tests -->
+                                        <artifact>io.quarkus:quarkus-platform-sample-app-test:${quarkus-platform-bom-generator.version}</artifact>
+                                        <maven-failsafe-plugin>true</maven-failsafe-plugin>
+                                        <systemProperties>
+                                            <java.util.logging.manager>org.jboss.logmanager.LogManager</java.util.logging.manager>
+                                            <quarkus.platform.group-id>${dollarSign}${dollarSign}{platform.groupId}</quarkus.platform.group-id>
+                                            <quarkus.platform.version>${dollarSign}${dollarSign}{platform.version}</quarkus.platform.version>
+                                            <maven.home>${dollarSign}${dollarSign}{maven.home}</maven.home>
+                                            <quarkus.platform.maven-plugin.pom>${dollarSign}${dollarSign}{project.file.absolutePath}</quarkus.platform.maven-plugin.pom>
```
aloubyansky, line 647: I should have added a comment, this option makes sure the POM that is used to resolve the Maven plugin artifact and its dependencies is the flattened version (`.flattened-pom.xml` produced by the flatten-maven-plugin when this module is built) instead of the original `pom.xml`.
So we run the tests using the actual artifacts that will be installed: the platform Maven plugin `quarkus-maven-plugin/target/quarkus-maven-plugin-<version>.jar`, `quarkus-maven-plugin/.flattened-pom.xml`, and the generated `quarkus-bom`.

aloubyansky: Pushed an update with a comment

PR #1341, opened 2024-11-06
Update to Blaze-Persistence 1.6.14

beikov:

Thanks!


gsmet:
Thanks! Will be part of 3.17!

PR #1362, opened 2024-11-28
Upgrade to the platform generator 0.0.117

aloubyansky:
Not a critical upgrade. This version of the generator is based on Quarkus 3.15.2.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -84,7 +84,7 @@
         <quarkus-vault.version>4.1.0</quarkus-vault.version>
         <quarkus-operator-sdk.version>6.9.1</quarkus-operator-sdk.version>
 
-        <quarkus-platform-bom-generator.version>0.0.116</quarkus-platform-bom-generator.version>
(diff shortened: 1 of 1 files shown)


pom.xml
```diff
@@ -84,7 +84,7 @@
         <quarkus-vault.version>4.1.0</quarkus-vault.version>
         <quarkus-operator-sdk.version>6.9.1</quarkus-operator-sdk.version>
 
-        <quarkus-platform-bom-generator.version>0.0.116</quarkus-platform-bom-generator.version>
+        <quarkus-platform-bom-generator.version>0.0.117</quarkus-platform-bom-generator.version>
```
gsmet, line 87: @aloubyansky do we need to copy the exclusions here or having them for the core BOM is enough?

aloubyansky: Having them in the core BOM is enough

gsmet (approved):

PR #1376, opened 2024-12-11
Upgrade to Quarkus 3.17.4

gsmet:

Thanks!


ppalaga:
Thanks for the heads up, @gsmet, indeed, I have seen this happening very rarely in the past with older versions of Quarkus, hence probably not a regression.

PR #1391, opened 2025-01-17
[3.15] Upgrade Camel Quarkus to 3.15.2

jamesnetherton:
This PR isn't addressing any specific compatibility concerns.

I know there's a process in place with branch freezes etc. But I was hoping we could get this merged instead of waiting until the next patch release in March.


jmartisk:
WDYT @gsmet?  I'm not sure we should be doing it this late and the changelog looks quite big...On the other hand, CI is passing... I'd do it but it's your decision I guess

gsmet:
Hey James,

As we are navigating through the new process and the discussions happened relatively late for this release cycle, we agreed with Jan that we could make an exception this time.

Just keep in time that we will enforce the process next time so make sure you schedule things appropriately.

We probably also need to reboot the discussion to reduce these delays so that it's more comfortable for everyone. More on that later.

Thanks!

PR #1450, opened 2025-03-28
Upgrade to Google Cloud Services extension pack 2.16.0

loicmathieu:
Can it be backported to 3.21 as it's based on 3.21, but it didn't make it in time for the release?


gsmet:
Thanks, I will prepare the backports a bit later.

loicmathieu:
Thanks!

PR #1513, opened 2025-06-12
Experiment with repository filters

gsmet:

Thanks!

### DIFF
--- a/.mvn/maven.config
+++ b/.mvn/maven.config
@@ -1,3 +1,6 @@
 -Dmaven.wagon.httpconnectionManager.ttlSeconds=120
 -Dmaven.wagon.http.retryHandler.requestSentEnabled=true
 -Dmaven.wagon.http.retryHandler.count=10
+-itr
+-Daether.remoteRepositoryFilter.groupId=true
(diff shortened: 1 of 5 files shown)


gsmet:
Failures are unrelated.

@aloubyansky what's your take on this?

My analysis is that:
- I expect all aligned dependencies that are listed here to be available in Maven Central
- Dependencies that are not in Maven Central will be listed explicitly IMO

This change doesn't affect the generated files so this seems fine - at least in its current state.

It might not be perfect but I think it's a lot better than having contributions from random repositories.

aloubyansky:
Isn't it supposed to have an effect on test dependencies? If repositories configured in transitive dependencies are disabled, how the Atlassian dependencies of `generated-platform-project/quarkus-camel/integration-tests/camel-quarkus-integration-test-jira` are resolved?

aloubyansky:
Ok, I still had the Atlassian repo enabled in the root POM, that's why.

I think the change is ok. We may need to mention in the docs how this is configured.

gsmet:
Will work on some doc soon.

PR #1526, opened 2025-07-09
Upgrade to Debezium 3.2.0.Final

Naros:


apupier:
the version of Hibernate used in Quarkus seems to be 7.x although Debezium is on 6.x
https://github.com/quarkusio/quarkus/blame/d5892635590dc4f832abf19b800968fea2a58e45/pom.xml#L74

the upgrade in Quarkus from 6 to 7 has been done with https://github.com/quarkusio/quarkus/commit/e391bea9bf43e4f2b57ddb24c287b9877eaf09b9#diff-9c5fb3d1b7e3b0f54bc5c4182965c4fe1f9023d449017cece3005d3f90e8e4d8 which means it arrived with Quarkus 3.24.0

apupier:
but the upgrade to Debezium 3.1 is already part of Quarkus 3.24.3 and it wasn't a problem https://github.com/quarkusio/quarkus-platform/commit/09aa94a59f358966008aa92be188e2438ded050f

jamesnetherton:
We should maybe consider to hold merge of this PR. It breaks the Camel Debezium components due to `NoClassDefFoundError: io/debezium/embedded/EmbeddedEngine`. We don't see it here because the itests are currently disabled (I forget why, I need to see if I can enable them again).

However, we're not in an LTS stream here, so I am somewhat more relaxed about it. But Camel Quarkus won't be in alignment with DBZ 3.2.0.Final until either CQ 3.26.0 or 3.27.0 is released (probably next month and beyond).

> but the upgrade to Debezium 3.1 is already part of Quarkus 3.24.3 and it wasn't a problem https://github.com/quarkusio/quarkus-platform/commit/09aa94a59f358966008aa92be188e2438ded050f

@apupier The Blaze Persistence test failure is already a known issue related to the Hibernate 7 upgrade & unrelated to this PR.

apupier:
>  It breaks the Camel Debezium components due to NoClassDefFoundError: io/debezium/embedded/EmbeddedEngine

On this part, I think we can update the Camel Debezium component as this class was deprecated for quite some time already even without upgrading Debezium to 3.2. it could allow a smoother transition, created https://github.com/apache/camel/pull/18582

Naros:
> the version of Hibernate used in Quarkus seems to be 7.x although Debezium is on 6.x https://github.com/quarkusio/quarkus/blame/d5892635590dc4f832abf19b800968fea2a58e45/pom.xml#L74

Ya we haven't had a chance to bump Hibernate to ORM 7. I'll make this a priority in early 3.3 so that we can look at bumping to 3.2.1.Final with backporting that upgrade here.

gsmet (approved):

gsmet:
The BlazeBit error is known and not related to Debezium.

As for Camel Quarkus, what's your preference you all, should we merge for 3.25 or wait?

jamesnetherton:
Then unless there's some urgency to have this change merged, lets wait until Camel catches up.

gsmet:
OK, then let's update to 3.2.1 when available and wait for the next round of releases with 3.26 to include this.

In any case, let's make sure this situation is sorted out before the next LTS (which will be 3.27).

gsmet:
@jamesnetherton I think we can wait.

gsmet:
Ah ah, I was going to ask you @jamesnetherton but then I thought "James is very thorough, check the issue", I wasn't disappointed :).

PR #1527, opened 2025-07-11
Update to Blaze-Persistence 1.6.16

beikov:

Thanks!


beikov:
Can I specify a profile to use somehow? It's a bit of a bummer, but Quarkus 3.24+, since it uses Hibernate ORM 7.0, will require a different integration. My testsuite project only includes that version if the `hibernate-7.0` profile is active though.

beikov:
Can I specify a profile to use somehow? It's a bit of a bummer, but Quarkus 3.24+, since it uses Hibernate ORM 7.0, will require a different integration. My testsuite project only includes that version if the `hibernate-7.0` profile is active though.

gsmet:
@beikov I tried to push something, let's see how it goes!

PR #1532, opened 2025-07-16
Upgrade to Quarkus 3.25.0.CR1

gsmet:

Thanks!

### DIFF
--- a/generated-platform-project/quarkus-camel/bom/pom.xml
+++ b/generated-platform-project/quarkus-camel/bom/pom.xml
@@ -8356,17 +8356,17 @@
       <dependency>
         <groupId>org.jetbrains.kotlin</groupId>
         <artifactId>kotlin-scripting-compiler-embeddable</artifactId>
-        <version>2.1.21</version>
+        <version>2.2.0</version>
       </dependency>
       <dependency>
         <groupId>org.jetbrains.kotlin</groupId>
         <artifactId>kotlin-scripting-compiler-impl-embeddable</artifactId>
-        <version>2.1.21</version>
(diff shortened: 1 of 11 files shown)


loicmathieu:
We had a bunch of firestore container issues lately, old version definitely suffer from it.
I'll have a look later as it fail in ecosystem CI also.

jamesnetherton:
> @jamesnetherton could you check what's going on with the Camel Quarkus failures?

I already took care of it https://github.com/apache/camel-quarkus/pull/7521.

We do nightly builds against the Quarkus `main` branch, so I am aware of these issues well in advance.

jamesnetherton:
> why you couldn't find the class anymore

Might be because `quarkus-devtools-utilities` got removed from `quarkus-junit5-internal`.

gsmet:
@loicmathieu we are using 2.17.1 already in `main` and we have the failure in `main`. From what I can see the container still requires Java 21.

This is going to be an issue as our LTS branches are failing and it's going to be hard to track their statuses.

Is there a way to pinpoint a version of the container that doesn't require Java 21?

If not, maybe we will need to update our CI to Java 21 but not checking our minimal version works is a bit unfortunate.

/cc @aloubyansky @rsvoboda

loicmathieu:
@gsmet sorry, the fix is not released yet.

I'll do a 2.17.2 release, older releases on LTS branches should not be affected as it's a rather new dev services.

gsmet:
@loicmathieu well, I think they are affected. I have the error in 3.15 and 3.20 AFAICS. But let me check...

PR #1540, opened 2025-07-24
Upgrade to Quarkiverse Google Cloud Services extension pack 2.18.0

loicmathieu:
@gsmet this new release is based on Quarkus 3.25 and contains the fix for the firestore container that fails tests.

Tell me if I need to create micro releases for older versions of Quarkus with this fix.


loicmathieu:
I'll prepare the micro versions soon

PR #1558, opened 2025-08-13
Upgrade to Quarkus 3.26.0.CR1

gsmet:

Thanks!


gsmet:
I'm surprised we don't have more tbh... I would have expected the `@ConfigMapping` lagging extensions to cause some failures.

PR #1893, opened 2026-04-16
Upgrade to platform generator 0.0.130

aloubyansky:
This release catches misconfigured and unresolvable relocations found in input BOMs.

### DIFF
--- a/generated-platform-project/quarkus-mcp-server/bom/pom.xml
+++ b/generated-platform-project/quarkus-mcp-server/bom/pom.xml
@@ -122,11 +122,6 @@
       </dependency>
       <dependency>
         <groupId>io.quarkiverse.mcp</groupId>
-        <artifactId>quarkus-mcp-server-sse-parent</artifactId>
-        <version>1.11.1</version>
-      </dependency>
-      <dependency>
-        <groupId>io.quarkiverse.mcp</groupId>
         <artifactId>quarkus-mcp-server-sse</artifactId>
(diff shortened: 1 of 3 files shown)


gsmet:
Is it something that should be fixed in the MCP Server project?

aloubyansky:
> Is it something that should be fixed in the MCP Server project?

Yes. The way the dependency appears in the BOM
```
      <dependency>
        <groupId>io.quarkiverse.mcp</groupId>
        <artifactId>quarkus-mcp-server-sse-parent</artifactId>
        <version>1.11.1</version>
      </dependency>
```
implies it's a JAR, while the artifact is supposed to be a POM.
The tricky thing about this specific case is
* it's a relocation POM
* that does not have `<packaging>pom</packaging>` (meaning it defaults to `<packaging>jar</packaging>`)
* `io.quarkiverse.mcp:quarkus-mcp-server-sse-parent:1.11.1` can be resolved as a JAR (empty/metadata-only JAR), if it is resolved directly
* but resolving `io.quarkiverse.mcp:quarkus-mcp-server-sse-parent:1.11.1` as a dependency (such as direct dependency in project `pom.xml`) will fail because the resolver will follow the relocation and the relocation is a POM-only artifact.

The MCP Server BOM is generated as part of the build. So either there has to be some config option to not add it or there is an issue in the BOM generation.

gsmet (approved):

PR #1982, opened 2026-05-25
Add quarkus-http-problem

### LINKED ISSUE #422: Add extension to quarkus-universe-bom
Not sure if/how feasible this is, but would be great to not ask developers to bump versions manualy

lwitkowski:
Core `dependencyManagement` only, no platform IT yet (integration-test not published as test-jar; will follow up). 

Closes [#422](https://github.com/quarkiverse/quarkus-http-problem/issues/422).


lwitkowski:
@gsmet would you mind taking a look? Would be awesome to include this for 4.0 :)

gsmet (commented):
Hey there, sorry for the delay.

I'm wondering why you didn't make this one a proper member? I'm asking because I would like the integration tests to run in the Platform to make sure it's compatible with the current Quarkus version.

lwitkowski:
> I'm wondering why you didn't make this one a proper member? I'm asking because I would like the integration tests to run in the Platform to make sure it's compatible with the current Quarkus version.

@gsmet that's because I have no clue what I'm doing here :). That's the first time I'm adding anything here, I couldn't find any guidelines, so it's rather reverse engineering from older PRs onboarding other extensions.

What do you mean 'proper member'? Is it only about missing integration tests? I saw some extensions don't have it but it was not a stopper to add them. `quarkus-http-problem` does not publish Integration Tests (yet). If that's a prerequisite - I'm more than happy to add this (or rather 'un-ignore' ITs from publishing).

aloubyansky:
@gsmet it doesn't have to be its own member to run tests. We have other extensions "injected" in the core bom with tests.

aloubyansky:
@lwitkowski we need integration tests to be published to be able to include the extension in the platform and run its testsuite to check compatibility.
The deployment is skipped here https://github.com/quarkiverse/quarkus-http-problem/blob/main/integration-test/pom.xml#L74
Could you have a look into that? Thanks.

lwitkowski:
Got it, I'm on it!
edit: we have there a few modules testing compatibility with optional deps (openapi, zalando problem lib etc) - those serve the very similar purpose - to check compatibility with others. I expect we need only our `core ITs`  for the purpose of platform. I will update this PR once those are published to maven central.

cescoffier:
@lwitkowski Anything we can do to help you on this?

gsmet:
Tests are now included! Let's push to CI.

PR #1985, opened 2026-05-29
Upgrade to Google Cloud Services 2.22.0

loicmathieu:
Please do a backport to Quarkus 3.35 as I think it would fix native issues (at least inside the Firestore extension) and align with Camel Quarkus dependency update for the Google Cloud Libraries.


jamesnetherton:
Probably best to remove the dependency overrides I added previously to preserve compatibility between members, while we were out of alignment with Google Cloud BOMs:

https://github.com/loicmathieu/quarkus-platform/blob/7f9db64470fa898a8c029387e8efff5babecb6b1/pom.xml#L404-L420

loicmathieu:
@jamesnetherton I'll do the change

jamesnetherton:
@loicmathieu Actually maybe the overrides would need to be kept for this PR. I see 2.21.0 still aligns to libraries-bom 26.76.0.

The dependency override removal would probably be better for #1987, because libraries-bom is bumped to 26.83.0.

loicmathieu:
@jamesnetherton sorry, the title was wrong (updated), it's an update to 2.22.0 which updates the Google Library Bom

loicmathieu:
@jamesnetherton I did the change, ready to GO!

gsmet (approved):

gsmet:
I tried to backport this but the Platform plugin is doing something extremely odd, downloading a gazillion of versions of each Google dependency and I had to abort.

gsmet:
I'm not sure what's going on but I had to revert this change as I was unable to release the CR1. It was downloading a gazillion of jars and not making any progress on the sync.

loicmathieu:
I had the same locally, no idea why...

gsmet:
@aloubyansky fixed it and I reapplied the changes this morning.

PR #2012, opened 2026-06-11
Upgrade to Quarkus 3.37.0.CR1 and Revert "Upgrade to Google Cloud Services 2.21.0"

gsmet:


ppalaga:
FWIW, there is QCXF 3.37.0.CR1 that should solve these issues.

PR #2185, opened 2026-09-17
Fix various semantic, technical, and grammatical errors in description for integration category

holly-cummins:
I noticed this when I was looking at something else, because 'skew' jumped out at me, and I suspected 'slew' was the intended word. 

But the more I looked, the more problems I saw :) 

- missing conjunction between 'read' and 'write'
- 'skew' should be 'slew', or, ideally, a less casual and more universally understood word
- S# is presumably a shift-key typo and should be S3
- Twitter no longer exists


aloubyansky:
@ppalaga @jamesnetherton fyi

ppalaga:
Thanks for the fix, @holly-cummins. Could please anybody explain where that description is shown? 

What does the `pinned` selection do? 

```
      "metadata": {
        "pinned": [
          "org.apache.camel.quarkus:camel-quarkus-core",
          "org.apache.camel.quarkus:camel-quarkus-core-xml"
        ]
      }
```

Please not that [camel-quarkus-core-xml](https://mvnrepository.com/artifact/org.apache.camel.quarkus/camel-quarkus-core-xml) does not exist since Camel Quarkus 1.0.0

holly-cummins:
> Thanks for the fix, @holly-cummins. Could please anybody explain where that description is shown?

It's present in the registry, and is shown, for example, in the json in https://registry.quarkus.io/q/swagger-ui/#/Client/get_client_categories_all. That's only visible to people who hit up the registry REST endpoint, so a limited audience. :) That's probably why no one noticed the issues with this description until now. 

However, as part of the more general revamp of categories to make them more usable for extension authors (https://github.com/quarkusio/quarkus/issues/55974), I'm planning to list our categories in the docs, in a table. That table would include the category description. 

> What does the `pinned` selection do?

I think "pinned" is saying that the name extensions would be attached to this category, even if the extension metadata says something different. 
> 
> ```
>       "metadata": {
>         "pinned": [
>           "org.apache.camel.quarkus:camel-quarkus-core",
>           "org.apache.camel.quarkus:camel-quarkus-core-xml"
>         ]
>       }
> ```
> 
> Please not that [camel-quarkus-core-xml](https://mvnrepository.com/artifact/org.apache.camel.quarkus/camel-quarkus-core-xml) does not exist since Camel Quarkus 1.0.0

I assume having stale extensions pinned is probably harmless, which is also why we haven't noticed this issue. But we should perhaps do some tidying here, I agree.

ppalaga (approved):
Thanks, sounds good, @holly-cummins

gsmet (approved):

